PR #3, opened 2020-10-30
create login signup pages

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/702429655/pulses/788451685)

## This PR
Created login and signup pages (and corresponding containers)
- Look consistent with designs
  - Design uses font-family: Circular, used sans-serif as the fallback (in screenshots)
- Functional links to login/signup pages (no forgot password form exists)
  - Added paths to App
- Use AntD 24 Grid System for layout flexibility/consistency

## Screenshots
Left: Design, Right: Finished Page (excluding NavBar)

## Verification Steps
- Compared side-by-side to see if the pages generally matched the design
- Shrunk/maximized window size to check flexibility
- Clicked links to check functionality
- Checked that color codes matched the design


src/containers/signup/Signup.tsx
```diff
@@ -0,0 +1,117 @@
...
+                                    >
+                                        <Input placeholder="Email"/>
+                                    </Form.Item>
+                                </Col>
+                                <Col span={rSpan}>
+                                    <Form.Item
+                                        name="cemail"
+                                        rules={[{ required: true, message: 'Please confirm your email!' }]}
+                                    >
+                                        <Input placeholder="Confirm Email"/>
```
florisdobber, line 60: We don't confirm email addresses. That's a mistake in the design sorry, but can you take it out and just make it a long input field like the password ones below?

florisdobber (changes requested):
Looks great overall! Just the minor comment

src/containers/signup/Signup.tsx
SofieDunt: Fixed!

jackblanc:
You should be able to fix the CI build with `npm run lint-fix && npm run prettier-fix` :)

jackblanc:
I resolved the merge conflicts from the changes to frontend-scaffold

src/containers/login/login.less
```diff
@@ -0,0 +1,78 @@
+.body-content-container {
```
jackblanc, line 1: 💭 Could we share some of these styles between the two containers? It looks like a lot of them are the same between the two pages.

jackblanc:
@SofieDunt I made a bunch of changes to frontend scaffold and wanted to make sure they work properly across downstream repos, so that's the reason for the commits above.

florisdobber (approved):
Looks good to me!

PR #8, opened 2020-11-04
Create page header

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/702429655/pulses/816228845)

## This PR

Made a page header component, styled according to designs
- page header div contains a title (defined by pageTitle prop) and a subtitle (defined by pageSubtitle prop)
- subtitle color is defined by textColor prop of StyledSubtitle component, which is defined by subtitleColor prop of PageHeader component
  - default color is '#444444'

## Screenshots
Created new page and added component
pageheaderex1
Added component to signup page
pageheaderex2

## Verification Steps

Imported and used the component in other pages and visually confirmed the page header matched the design (font, color, size, etc)
- import PageHeader from '../../components/pageheader/PageHeader'
- tested various subtitleColor options


florisdobber (commented):
Looks good to me! Jack might have some comments on the code

src/components/pageheader/PageHeader.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import './pageheader.less';
+import { Typography } from 'antd';
+const { Title, Paragraph } = Typography;
+
+type PageHeaderProps = {
```
jackblanc, line 6: Prefer to use interfaces for props 
```suggestion
interface PageHeaderProps {
```

src/components/pageheader/PageHeader.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const PageHeader: React.FC<PageHeaderProps> = ({
+  pageTitle,
+  pageSubtitle,
+  subtitleColor,
+}) => {
+  return (
+    <div className="header">
+      <Title id="title">{pageTitle}</Title>
+      <Paragraph className="subtitle" id={subtitleColor}>
```
jackblanc, line 20: I think you should be able to use className here instead of id, perhaps something like 
```suggestion
      <Paragraph className={`subtitle ${subtitleColor}`}>
```

src/components/pageheader/PageHeader.tsx
```diff
@@ -0,0 +1,27 @@
...
+};
+
+const PageHeader: React.FC<PageHeaderProps> = ({
+  pageTitle,
+  pageSubtitle,
+  subtitleColor,
+}) => {
+  return (
+    <div className="header">
+      <Title id="title">{pageTitle}</Title>
```
jackblanc, line 19: Why can't className be used for this?

SofieDunt: Using Title, I wasn't able to style certain things with className but I've fixed that with !important (same with Paragraph and subtitleColor above).

jackblanc: Unfortunately, using `!important` can cause a lot of unintended side effects, especially when we're working with the Ant.D library -`!important` has the potential to override AntD's internal styles in unpredictable ways. 

A good way to properly overwrite the styles without using `!important` is using styled components. There is some WIP documentation in https://github.com/Code-4-Community/c4c-dev-docs/pull/12/files. 

Ultimately, you'll have something like: 

```
const StyledTitle = styled(Title)`
   your styles
`;

<StyledTitle>{pageTitle}</StyledTitle>
```

I believe the dependencies are not installed currently, you can install them following the guide in the docs. Let me know if you have any questions on this :) 

cc @rymaju

src/components/pageheader/PageHeader.tsx
```diff
@@ -0,0 +1,46 @@
...
+  color: #3a681a;
+  font-weight: bold;
+  margin-bottom: 0px;
+`;
+
+const StyledSubtitle = styled(Paragraph)`
+  font-weight: normal;
+  font-size: 24px;
+  line-height: 32px;
+  color: ${(props: StyledSubtitleProps) =>
```
jackblanc, line 29: ❓ Do we want the header color to be customizable? It's the same on all pages on Figma

src/components/pageheader/PageHeader.tsx
```diff
@@ -0,0 +1,46 @@
...
+    props.textColor ? props.textColor : '#444444'};
```
jackblanc, line 30: 💭 You can add `#444444` to `src/colors.tsx` for extensibility

src/components/pageheader/pageheader.less
```diff
@@ -0,0 +1,4 @@
+.header {
+  font-family: Circular, sans-serif;
+  font-style: normal;
+}
```
jackblanc, line 4: 🤔 Font family should get set globally in Ant.D config, we might have to update it for this project if it's not correct.

jackblanc (commented):
Great work with the styled components! Left a few questions 👍

src/components/pageheader/PageHeader.tsx
SofieDunt: The main header color is constant but the subheader color has two different options on Figma (#444444 for logged-in landing and #3A681A for the leaderboard pages)

src/components/pageheader/pageheader.less
SofieDunt: That might be part of the global styles ticket - I'm not sure. Should I take this out now and assume the config will get updated later?

jackblanc: Yeah, it'll get updated in global styles, so you can take this out 👍

jackblanc (approved):
Looks good to me! To get Travis to pass, try running `npm run prettier-fix`

PR #14, opened 2020-11-24
Create settings page

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/702429655/pulses/816231284)

Settings page for users to edit their account settings or delete/deactivate their account, currently just styling

## This PR

Updated Settings.tsx and created settings.less to match the designs
- Used styled components for form titles and submit buttons to override internal styles
- Used Col grids to space out the form input fields

## Screenshots

Settings

## Verification Steps

- Visually checked that the page matches the designs
- Shrunk/maximized window size to check flexibility


src/containers/settings/Settings.tsx
```diff
@@ -12,17 +38,116 @@ const Settings: React.FC = () => {
...
+                <Col span={formHalfItemSpan}>
+                  <Form.Item name="last-name">
+                    <Input placeholder="Last Name" />
+                  </Form.Item>
+                </Col>
+              </Row>
+              <Form.Item {...formLayout} name="username">
+                <Input placeholder="Username" />
+              </Form.Item>
+              <Form.Item {...formLayout} name="phone-number">
```
florisdobber, line 73: Sorry the designs aren't accurate in this but there's no phone number field.

src/containers/settings/Settings.tsx
```diff
@@ -12,17 +38,116 @@ const Settings: React.FC = () => {
...
+                rules={[
+                  {
+                    required: true,
+                    message: 'Please input your current password!',
+                  },
+                ]}
+              >
+                <Input placeholder="Current Password" />
+              </Form.Item>
+              <Form.Item
```
florisdobber, line 112: Also not in the designs, but could you put text below the new password field saying "Your new password must be at least 8 characters". I think there's the option to put something below a field on a form field.

florisdobber (changes requested):
Code looks great, just those two things which you couldn't have known about

src/containers/settings/Settings.tsx
SofieDunt: I also added a rule to require having at least 8 characters for the new password input.
UpdatedSettings 
PasswordRule

florisdobber: Amazing!!

florisdobber (approved):
Looks good, thanks for making those changes!

src/containers/settings/settings.less
```diff
@@ -0,0 +1,5 @@
+.page-content-container {
```
jackblanc, line 1: Is there any reason not to use a styled div for this? The same way we use other styled components. I think that'd be better for consistency.

src/containers/settings/Settings.tsx
```diff
@@ -12,17 +40,118 @@ const Settings: React.FC = () => {
     <>
       <Helmet>
         <title>Settings</title>
-        <meta name="description" content="Description goes here." />
+        <meta
+          name="Settings"
+          content="The user's account settings page, where they can change their profile information and password or deactivate/delete their account."
+        />
       </Helmet>
```
jackblanc, line 47: This page is going to be protected, so we don't need Helmet for SEO

jackblanc (approved):
Looks great! Left some comments with room for improvement

src/containers/settings/settings.less
SofieDunt: No, I was just only using styled components when necessary to override internal styles. It's now updated to use a styled div.

PR #20, opened 2020-12-14
Leaderboard

willmt80:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/702429655/views/12631042/pulses/816228893)

On the team and user leaderboard pages, this leaderboard allows users to see a ranked list of teams or users. There is customizability with the number of items per page, the number of tabs, the functionality of the right side of an item, and the collapsible content

## This PR

The leaderboard can be seen at the /dummy route. The size and spacing will be changed within the context of where the leaderboard is used so right now there is no styling on its width.

I added a few items to the antd global styling. As you can see in the screenshot, the coloring on the pagination items is not quite right. I am trying to avoid overriding antd styles and I have not been able to figure out the name of that specific item. I changed the primary color to lightgreen because all of the text colors were defaulting to blue.

## Screenshots

## Verification Steps

I have a couple of examples of the kinds of collapse panels in the leaderboard, made the dummy page to check that interactions were correct. Current usage only has one tab.

No collapse

with collapse

page 2


src/App.tsx
```diff
@@ -35,6 +36,7 @@ const App: React.FC = () => {
               <Route path="/signup" exact component={Signup} />
               <Route path="/home" exact component={Home} />
               <Route path="/settings" exact component={Settings} />
+              <Route path="/dummy" exact component={DummyLeaderboard} />
```
jackblanc, line 39: Rather than adding this `dummy` leaderboard and having to change things later, I think it makes more sense to initialize the proper container now. You can make this one the `VolunteerLeaderboard` as the `TeamLeaderboard` will come later in Milestone 2.

src/components/leaderboard/leaderboard-tab/LeaderboardTab.tsx
```diff
@@ -0,0 +1,104 @@
...
+const LeaderboardItemRank = styled(Paragraph)<ParagraphProps>`
+  color: ${BLACK};
+  display: inline;
+  padding-right: 20px;
+  vertical-align: middle;
+  font-size: 24px;
+  line-height: 0px;
+`;
+
+type LeaderboardTabProps = {
```
jackblanc, line 37: I think this makes more sense as an `interface` rather than a `type`.

src/components/leaderboard/leaderboard-tab/LeaderboardTab.tsx
```diff
@@ -0,0 +1,104 @@
...
+  tabItems: LeaderboardItem[];
+  currentPage: number;
+  itemsPerPage: number;
+};
+
+export type LeaderboardItem = {
```
jackblanc, line 43: `interface`

src/components/leaderboard/leaderboard-tab/LeaderboardTab.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+const LeaderboardTab: React.FC<LeaderboardTabProps> = ({
+  tabItems,
+  currentPage,
+  itemsPerPage,
+}) => {
+  const itemsOnPage = tabItems.slice(
+    (currentPage - 1) * itemsPerPage,
+    currentPage * itemsPerPage,
+  );
```
jackblanc, line 58: This is something we should unit test using Jest/Enzyme. I can pair on that.

src/components/leaderboard/leaderboard-tab/LeaderboardTab.tsx
```diff
@@ -0,0 +1,104 @@
...
+      </Row>
+    </>
+  );
+};
+
+/**
+ * Creates the antd Panel objects from the given items
+ * @param items
+ */
+function getPanels(items: LeaderboardItem[]): React.ReactNode {
```
jackblanc, line 75: Typically I have been taught that it's a better practice to leave all JSX in the return statement rather than separating it out into a function, which is called subrendering. I think if it warrants being separated out it should be a component, otherwise just include this map in the return statement above.

src/components/leaderboard/leaderboard-tab/LeaderboardTab.tsx
```diff
@@ -0,0 +1,104 @@
...
+          >
+            {item.collapseContent}
+          </LeaderboardPanel>
+        );
+      })}
+    </Collapse>
+  );
+}
+
+function getPanelHeader(name: string, rank?: number): React.ReactNode {
```
jackblanc, line 95: Same as above

src/components/leaderboard/leaderboard-tabs/LeaderboardTabs.tsx
```diff
@@ -0,0 +1,58 @@
+import * as React from 'react';
+import { Typography, Tabs, Pagination } from 'antd';
+import LeaderboardTab, {
+  LeaderboardItem,
+} from '../leaderboard-tab/LeaderboardTab';
+
+export type TabInfo = {
```
jackblanc, line 7: interfaces

src/components/leaderboard/leaderboard-tabs/LeaderboardTabs.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const onChangePage = (page: number): void => setCurrentPage(page);
+  const onChangeTab = (tab: string): void => setCurrentTab(tab);
+
+  return (
+    <Tabs
+      type="card"
+      defaultActiveKey={currentTab}
+      onChange={onChangeTab}
+      tabBarExtraContent={
+        <Pagination
```
jackblanc, line 35: The colors don't quite work here
https://user-images.githubusercontent.com/15748593/104765558-edc96900-571d-11eb-9bc8-6f001916d3ad.mov

src/containers/dummy-leaderboard/DummyLeaderboard.tsx
```diff
@@ -0,0 +1,49 @@
...
+      },
+      {
+        name: 'Will',
+        rightSide: 7,
+      },
+    ],
+  },
+];
+
+const DummyLeaderboard: React.FC = () => {
```
jackblanc, line 45: Let's rename this to `VolunteerLeaderboard` so we don't have to change it later.

jackblanc (commented):
This looks great! Left a few comments, please reach out if you have any questions! There are also a few things off with the Figma, take a look at the Hi-Fi 2 page for the most up-to-date styles.

willmt80:
Made most of the changes requested in the comments, however I am still running into trouble with the coloring for the pagination tabs. The first tab is highlighted by default (as it is in all the examples here https://ant.design/components/pagination/) with the hover color but clicking any tab changes it to a different border/text color. Lots of trial and error with different color values and still not sure how to fix.

willmt80:
leaderboard_closed

leaderboard_open

florisdobber (commented):
Looks good generally, great job! See if you can get at least most of the colors right.

Noticed the dummy data includes rank. The back end does not return the data like that, just as a sorted list. @jackblanc is it better to return the data with a rank from the back end or can we add that on the front end?

jackblanc:
@florisdobber We can come up with rank on the frontend, that's a good point. We should just use the index value in the map to generate that. 

For the button colors, @rymaju may be able to help? let's ticket it separately if it takes too long. We can override manually if the issue is with configuring global variables?

jackblanc (approved):

PR #22, opened 2020-12-24
Optimize login/signup for mobile

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/925294541)

Makes the log in and signup pages easier to navigate on mobile devices/narrow windows

## This PR

- Updated to use styled components only
- In the Login and Signup components, only displays the form (no input or greeting container) if mobile
  - Created a SignupForm and LoginForm component, form layouts are dependent on whether the browser is mobile (rather than making mobile page components so the login/signup form only exists in one place)

## Screenshots

Login
LoginDesktop LoginMobile 

Signup
SignupDesktop SignupMobile

## Verification Steps

Visually confirmed the layout updated when resizing the browser, used Device Mode in Chrome Dev Tools to make sure the layout worked for mobile.


florisdobber (commented):
@jackblanc Code looks good to me and I ran it locally to check if it behaves as expected but it's a fair amount of code and I don't want to miss anything.

src/components/greeting-container/GreetingContainer.tsx
```diff
@@ -17,11 +17,11 @@ const InfoContainer = styled.div`
   background: url(${logo}) no-repeat bottom right #d4edaa;
   box-shadow: 1.58105px 3.16211px 6.32421px rgba(0, 0, 0, 0.09);
```
rymaju, line 18: Not in this PR but why are we using float values for pixels? Seems a little funky

src/components/login-form/LoginForm.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+const LoginButton = styled(Button)`
+  width: 96px;
+  margin-top: 20px;
+`;
+
+const LoginForm: React.FC<LoginFormProps> = ({ onFinish }) => {
+  const { windowType } = useWindowDimensions();
+
+  if (windowType === WindowTypes.Mobile) {
```
rymaju, line 27: Very interesting! Can you explain why you chose to use window dimensions as opposed to CSS media queries? I'll have to test this myself.

src/components/login-form/LoginForm.tsx
```diff
@@ -0,0 +1,111 @@
...
+                  message: 'Please input your password!',
+                },
+              ]}
+            >
+              <Input.Password placeholder="Password" />
+            </Form.Item>
+          </Col>
+        </Row>
+
+        <Form.Item style={{ marginBottom: '10px' }}>
```
rymaju, line 101: @florisdobber don't know how picky you want to be with styling, I would say try to keep everything as a styled-component but 🤷 might be a little excessive for something small like this.

src/components/signup-form/SignupForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+              ]}
+            >
+              <Input placeholder="First Name" />
+            </Form.Item>
+          </Col>
+          <Col
+            span={formHalfItemSpan}
+            style={{
+              marginLeft: `${
+                isMobile ? `${width > 575 ? '-7.5%' : '8%'}` : '-7.5%'
```
rymaju, line 97: This kind of inline styling seems a little too complex, can this be pulled out?

src/containers/signup/Signup.tsx
```diff
@@ -1,30 +1,44 @@
...
+} from '../../components/window-dimensions';
 
-const { Paragraph, Title } = Typography;
+const { Title } = Typography;
 
-const Footer: typeof Paragraph = styled(Paragraph)<ParagraphProps>`
-  color: ${TEXT_GREY};
-  line-height: 1.5;
+const SignupContainer = styled.div`
+  padding: 8%;
```
rymaju, line 20: ⛏️ Padding 8% seems a little odd? Usually padding is fixed, usually you dont want something that changes the layout of the whole screen to be dyanmic like this because it makes the browser do a lot of redraws.

src/containers/signup/Signup.tsx
```diff
@@ -1,30 +1,44 @@
...
+  min-width: 250px;
+  padding: 30px 20px 20px 50px;
+  background: ${LIGHT_GREY};
+  box-shadow: 2px 3px 6px rgba(0, 0, 0, 0.09);
+  border-radius: 6px;
+`;
+
+const Line = styled.div`
+  height: 2px;
+  margin: 5% -20px 7% -50px;
```
rymaju, line 34: Same as above

rymaju (commented):
Ok so the biggest thing I'm gonna look into is how youre doing mobile vs desktop styling. It looks like youre doing it programmatically with javascript, I *think* its better to do it with CSS media queries, I'll comment again after I do some research. Other than that looks solid! 👍

src/components/greeting-container/GreetingContainer.tsx
SofieDunt: I think it was just copy-pasted from the Figma design

src/components/login-form/LoginForm.tsx
SofieDunt: See Slack

src/components/signup-form/SignupForm.tsx
SofieDunt: Yeah, the logic can be done earlier but not separately in a styled component because it's using the `useWindowDimensions` hook

src/containers/signup/Signup.tsx
SofieDunt: You're right, I can fix this - originally did it this way because the padding needs to change for the different window types and this seemed a bit easier than using the hook

src/components/login-form/LoginForm.tsx
jackblanc, line 27: In this case, I think a `switch / case` statement would be more readable. It explicitly shows which types correspond to which code.

src/components/login-form/LoginForm.tsx
jackblanc: Yeah let's keep this in a styled-component, I think future people working on the codebase would appreciate all of the styles being in one consistent location.

src/components/signup-form/SignupForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+      Sign Up
+    </Button>
+  );
+  const FooterToLogin = (
+    <Footer>
+      ALREADY HAVE AN ACCOUNT?
+      <br />
+      LOGIN <Link to="/login">HERE!</Link>
+    </Footer>
+  );
```
jackblanc, line 45: Typically we should avoid having JSX outside of the return statement of a component, I think it makes more sense here to include these in the return, duplicated code is less of an issue.

src/components/signup-form/SignupForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+interface SignupFormProps {
+  readonly onFinish: any;
+}
+
+interface ButtonFooterProps {
+  readonly isMobile: boolean;
+}
+
+const ButtonFooter: React.FC<ButtonFooterProps> = ({ isMobile }) => {
```
jackblanc, line 33: If this is large enough to be a separate component, it should go in it's own file as well. Personally, I think it makes more sense to include it directly on line 137.

src/components/signup-form/SignupForm.tsx
jackblanc: +1

src/containers/login/Login.tsx
```diff
@@ -9,103 +8,112 @@ import { login } from '../../auth/ducks/thunks';
...
+
+      <Footer>
+        NEW TO SPEAK FOR THE TREES?
+        <br />
+        SIGN UP <Link to="/signup">HERE!</Link>
+      </Footer>
+    </div>
+  );
+
+  if (windowType === WindowTypes.Mobile) {
```
jackblanc, line 72: `switch / case` here as well.

src/containers/login/Login.tsx
```diff
@@ -9,103 +8,112 @@ import { login } from '../../auth/ducks/thunks';
...
-            </Paragraph>
-
-            <Footer>
-              NEW TO SPEAK FOR THE TREES?
-              <br />
-              SIGN UP <Link to="/signup">HERE!</Link>
-            </Footer>
-          </Col>
-          <Col span={2}></Col>
+          <Col span={`${width < 715 ? 1 : 2}`}></Col>
```
jackblanc, line 104: Can we add a constant for this `715`? What does it represent?

src/containers/signup/Signup.tsx
```diff
@@ -39,114 +53,37 @@ const Signup: React.FC<SignupProps> = ({ tokens }) => {
 
   const greetingHeader = 'Welcome Back!';
 
+  if (windowType === WindowTypes.Mobile) {
```
jackblanc, line 56: switch / case

src/containers/signup/Signup.tsx
```diff
@@ -1,30 +1,44 @@
...
+  WindowTypes,
+} from '../../components/window-dimensions';
 
-const { Paragraph, Title } = Typography;
+const { Title } = Typography;
 
-const Footer: typeof Paragraph = styled(Paragraph)<ParagraphProps>`
-  color: ${TEXT_GREY};
-  line-height: 1.5;
+const SignupContainer = styled.div`
```
jackblanc, line 19: `SignupContainer`, and Login, are a bit misleading to read. We're defining the SignupContainer in this file as a React component, so there is a duplicated meaning.

src/containers/signup/Signup.tsx
```diff
@@ -39,114 +53,37 @@ const Signup: React.FC<SignupProps> = ({ tokens }) => {
...
-                </Col>
-              </Row>
-            </Form>
-          </Col>
+            <Line />
+            <SignupForm onFinish={onFinish} />
+          </InputContainer>
 
-          <Col span={2}></Col>
+          <Col span={`${width < 715 ? 1 : 2}`}></Col>
```
jackblanc, line 86: constant

jackblanc (commented):
This looks great! Left a few additional suggestions. Let me know if you have any questions or want to pair!

src/components/login-form/LoginForm.tsx
SofieDunt: Fixed

src/containers/login/Login.tsx
SofieDunt: It's a breakpoint for the width - if `width < 715` an offset span of `2` is too large. Added constant `OffsetSpanBreakpoint`

src/containers/signup/Signup.tsx
SofieDunt: Changed to `SignupPageContainer`, `MobileSignupPageContainer`, `LoginPageContainer`, etc.

src/components/login-form/LoginForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+  margin-bottom: 10px;
+`;
+
+const LoginForm: React.FC<LoginFormProps> = ({ onFinish }) => {
+  const { windowType } = useWindowDimensions();
+
+  switch (windowType) {
+    case WindowTypes.Mobile:
+      return (
+        <>
```
jackblanc, line 34: I don't think you should need these fragments here 🤔 
```suggestion
```

src/components/login-form/LoginForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+              <LoginButton type="primary" htmlType="submit" size="large">
+                Log In
+              </LoginButton>
+            </Form.Item>
+          </Form>
+        </>
+      );
+    default:
+      return (
+        <>
```
jackblanc, line 74: Fragment

src/components/signup-form/SignupForm.tsx
```diff
@@ -0,0 +1,141 @@
...
+          rules={[
+            {
+              required: true,
+              message: 'Please confirm your password!',
+            },
+          ]}
+        >
+          <Input.Password placeholder="Confirm Password" />
+        </Form.Item>
+        <div>
```
jackblanc, line 93: This is likely a place where you _would_ want the fragment, right? Or does this div affect styling?

jackblanc (approved):
Looks great :)

PR #24, opened 2021-01-23
reservation modal attempt

vrushalitarte01:
## Why

--
-handleOk and handleCancel have been moved so they can be added to containers
-set visibility also moved so can be added to containers

## This PR

--
Implemented requested changes, just needs final look over hopefully.

## Screenshots

--
block reserved by user:
Screen Shot 2021-02-07 at 3 02 26 PM
block reserved by another user:
Screen Shot 2021-02-07 at 3 04 21 PM
block available to reserve:
Screen Shot 2021-02-07 at 3 03 44 PM

## Verification Steps

--
To verify my changes work, I imported my modal in the login file and made a mock modal in the login page and would change its fields to mimic the different scenarios of blocks to ensure the modal that popup is as expected.


florisdobber (dismissed):
Can you change the style of the modal so it more closely resembles the design? So for release there's a red icon at the start and then a green check mark when the user does it successfully. Similar thing for reserve. 

Also make the modal appears when you click on something. You're currently displaying two modals. The one you're returning and the one you're creating use the functions.

src/components/ReservationModal.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import styled from 'styled-components';
+import { WHITE, BLACK } from '../colors';
+import { Modal } from 'antd';
+
+interface ReservationModalProps {
+    readonly status: string;
```
jackblanc, line 7:
```suggestion
enum ReservationModalType {
  OPEN = 'OPEN'
}

interface ReservationModalProps {
    readonly status: ReservationModalType;
```

You can make an enum to restrict which types are allowed for status.

src/components/ReservationModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+    readonly blockID: number;
+}
```
jackblanc, line 9: You should add three properties here: 
```
readonly onOk: () => void;
readonly onCancel: () => void;
readonly isVisible: boolean;
```
That way we can respond to the user's response with custom code. 
And we'll control visibility from the parent component.

src/components/ReservationModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+    });
+  }
+
+function info(blockID: number) {
+    StyledModal.warning({
+        title: 'Sorry',
+        content: 'Block ' + blockID + ' is not available to reserve/release',
+    });
+  }
+
```
jackblanc, line 51: I think a good way to do this would be to create a function which returns the content the modal requires. 

```
const getModalCopy = (status, blockId) => {
  switch (status) {
    case OPEN
```
Have this function return an object containing the copy.

Then, we'll use that content below.

src/components/ReservationModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+    if (status === 'open') {
+        warning(blockID);
+    }
+
+    if (status === 'taken') {
+        info(blockID);
+    }
+
+    return (
+        <StyledModal></StyledModal>
```
jackblanc, line 67:
```suggestion
        <StyledModal visible={visible} onCancel={onCancel} onOk={onOk} title={copy.title} content={copy.content} />
```

We only want to return one modal and provide it with the required props.

jackblanc (commented):
Off to a good start here! I left some suggestions, the code snippets I wrote aren't exact but they should be pretty close.

jackblanc:
Here's another modal example from the LLB project. 
https://github.com/Code-4-Community/lucys-love-bus-frontend/pull/4/files

src/components/ReservationModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+const StyledModal = styled(Modal)`
+    height: 40px;
+    background: ${WHITE};
+    color: ${BLACK};
+    font-size: 5px;
+    font-weight: 400;
+`;
+
+const ReservationModal: React.FC<ReservationModalProps> = ({ 
+  status, blockID, onOk, onCancel, isVisible}) => {
```
jackblanc, line 29: This formatting doesn't look right 🤔

src/components/ReservationModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+    const [visibility, setVisibility] = useState(isVisible);
```
jackblanc, line 30: I don't think that we want to store this in local component state, we should just directly use `isVisible` prop and hold the state in the parent component.

src/components/ReservationModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    const handleCancel = (): void => {
+      onCancel();
+      setVisibility(false);
+    }
+
+    const handleOk = (): void => {
+      onOk();
+      setVisibility(false);
+    }
```
jackblanc, line 51: These can be removed and we can have the modal below directly use onOk and onCancel after removing the local state of this component.

src/components/ReservationModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+          return (
+              <text>
+                <b>{`Sorry!`}</b><br/ > {`Block ${blockID} is not available to reserve/release`}
+              </text>
+          );
+      }
+    };
+
+    return (
+      <div>
```
jackblanc, line 77: This div can be removed

jackblanc (commented):
Looks like this is almost there, left a comment about removing the local state of this component.

src/components/ReservationModal.tsx
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
+import styled from 'styled-components';
+import { WHITE, BLACK } from '../colors';
```
jackblanc, line 3:
```suggestion
import { WHITE, BLACK } from '../utils/colors';
```
This file was moved in a recent commit, so it has to be updated.

jackblanc (approved):
Looks good, pending changing the file location. Nice work iterating on this!

src/components/ReservationModal.tsx
jackblanc: I committed the suggestion using GitHub's UI, the build should pass now :)

PR #25, opened 2021-01-25
Connect Leaderboard Routes

willmt80:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/925416107)

Connecting the leaderboard routes populates the leaderboard objects on the volunteer and teams pages with our data, allowing the user to see how who has logged the most trees

## This PR

Some of the features looked like they could be abstracted, such as the actions and a lot of the types that the leaderboards use. These are in the actions.ts file and types.ts file in the components/leaderboard/ducks directory. The other ducks files are in the respective container directories for the volunteer and teams leaderboards.

One of the features of the leaderboard pages is the ability to view the best volunteers/teams of the last week/month/year/all-time. Each of these possible time scales requires a different object in the store and each has its own reducer. I think it's possible that a ton of the code here can be abstracted, however I was having a hard time doing so.

## Screenshots

## Verification Steps

Unfortunately, it turns out that the axios library we are using cannot accept arguments in request bodies for get requests. This means we are unable to test that the routes pull in data properly.


florisdobber (commented):
Will review this in more detail when the back end is fixed. Should be soon

src/containers/team-leaderboard/TeamLeaderboard.tsx
```diff
@@ -0,0 +1,136 @@
...
+      <PageHeader
+        pageTitle="Team Leaderboard"
+        pageSubtitle="Let’s celebrate your team reaching for its goals!"
+      />
+
+      <LeaderboardContainer>
+        {teamLeaderboardItemsWeekly.kind === AsyncRequestKinds.Completed &&
+          teamLeaderboardItemsMonthly.kind === AsyncRequestKinds.Completed &&
+          teamLeaderboardItemsYearly.kind === AsyncRequestKinds.Completed &&
+          teamLeaderboardItemsAllTime.kind === AsyncRequestKinds.Completed && (
```
jackblanc, line 112: Just pointing out, this implies that all four variations of leaderboard must be loaded in order to view one of them. 

@florisdobber / @liammoyn what does the backend for this do when given a number other than 7, 30, etc? If it's truly flexible I think we should make this code a lot smaller.

src/components/leaderboard/leaderboard-tab/LeaderboardTab.tsx
```diff
@@ -42,19 +43,32 @@ const LeaderboardItemRank = styled(Paragraph)<ParagraphProps>`
...
+): TabItem[] => {
+  return leaderboardItems.map((item) => {
+    return {
+      id: item.id,
+      name: item.name,
+      rightSide: item.blocksCounted,
+    };
+  });
+};
+
```
jackblanc, line 71: Rather than putting the responsibility on the caller of this component to provide the properly formatted data, I think we should instead have this component take a list of `LeaderboardItem` and internally shape the data as needed. That way, we don't have to store dual states in the parent container.

src/containers/team-leaderboard/TeamLeaderboard.tsx
```diff
@@ -0,0 +1,136 @@
...
+    content: [],
+  });
+  const [yearlyTab, setYearlyTab] = useState<TabInfo>({
+    name: 'Yearly',
+    content: [],
+  });
+  const [allTimeTab, setAllTimeTab] = useState<TabInfo>({
+    name: 'All Time',
+    content: [],
+  });
```
jackblanc, line 60: Using React State in combination with Redux will get messy quickly. Related to my comment in the leaderboard component, we should pass the data directly into the component from the Redux store, which removes almost > 50 lines from this file.

src/containers/team-leaderboard/TeamLeaderboard.tsx
```diff
@@ -0,0 +1,136 @@
...
+    dispatch(getTeamsLeaderboard(LeaderboardPreviousDays.allTime));
+  }, []);
+
+  useEffect(() => {
+    teamLeaderboardItemsWeekly.kind === AsyncRequestKinds.Completed &&
+      setWeeklyTab({
+        ...weeklyTab,
+        content: leaderboardItemsToTabItems(teamLeaderboardItemsWeekly.result),
+      });
+  }, [teamLeaderboardItemsWeekly]);
```
jackblanc, line 75: To some extent, this defeats the purpose of using Redux at all.

src/api/apiClient.ts
```diff
@@ -0,0 +1,141 @@
...
+
+const getUsersLeaderboard = (
+  previousDays: number,
+): Promise<LeaderboardItem[]> => {
+  // return AppAxiosInstance.get(ApiClientRoutes.USERS_LEADERBOARD, {
+  //   params: {
+  //     previousDays,
+  //   },
+  // }).then((response) => response.data.users);
+  switch(previousDays) {
```
liammoyn, line 30: Seems like this switch statement can get thrown into a helper function pretty easily

src/containers/team-leaderboard/TeamLeaderboard.tsx
```diff
@@ -0,0 +1,136 @@
...
+      <LeaderboardContainer>
+        {teamLeaderboardItemsWeekly.kind === AsyncRequestKinds.Completed &&
+          teamLeaderboardItemsMonthly.kind === AsyncRequestKinds.Completed &&
+          teamLeaderboardItemsYearly.kind === AsyncRequestKinds.Completed &&
+          teamLeaderboardItemsAllTime.kind === AsyncRequestKinds.Completed && (
+            <LeaderboardTabs
+              tabs={[weeklyTab, monthlyTab, yearlyTab, allTimeTab]}
+              itemsPerPage={4}
+            />
+          )}
```
liammoyn, line 117: We should have some way of displaying an error if one of the calls fails right?

src/store.ts
```diff
@@ -15,31 +15,94 @@ import userReducer, { initialUserState } from './auth/ducks/reducers';
...
 export const initialStoreState: C4CState = {
   authenticationState: initialUserState,
+  userLeaderboardWeeklyState: initialUserLeaderboardWeeklyState,
+  userLeaderboardMonthlyState: initialUserLeaderboardMonthlyState,
+  userLeaderboardYearlyState: initialUserLeaderboardYearlyState,
+  userLeaderboardAllTimeState: initialUserLeaderboardAllTimeState,
+  teamLeaderboardWeeklyState: initialTeamLeaderboardWeeklyState,
+  teamLeaderboardMonthlyState: initialTeamLeaderboardMonthlyState,
+  teamLeaderboardYearlyState: initialTeamLeaderboardYearlyState,
+  teamLeaderboardAllTimeState: initialTeamLeaderboardAllTimeState,
```
jackblanc, line 99: Could you explain the reasoning for having a separate piece of state for each time period? In looking over this I see how it would be a bit confusing, as no reducers have more than one piece of state to them. 

I think there are two options here: 
- have a `userLeaderboardState` containing four sub-pieces of state, but all in one reducer. Store each slice of data within that reducer
- have a single `userLeaderboardState` that just stores one slice of data at a time. When a user changes the time period, we simply make a new request to the backend, and update the ONE piece of state to contain data for that time period only. 


Personally, I think option 2 is the cleanest solution. We don't have to worry about the data taking ~1 second to reload each time IMO. If implemented properly, we can just make the leaderboard capable of displaying _any_ range of data from today until X days in the past, though the UI doesn't have to reveal that.

src/store.ts
```diff
@@ -15,31 +15,94 @@ import userReducer, { initialUserState } from './auth/ducks/reducers';
...
 export interface C4CState {
   authenticationState: UserAuthenticationReducerState;
+  userLeaderboardWeeklyState: UserLeaderboardWeeklyReducerState;
+  userLeaderboardMonthlyState: UserLeaderboardMonthlyReducerState;
+  userLeaderboardYearlyState: UserLeaderboardYearlyReducerState;
+  userLeaderboardAllTimeState: UserLeaderboardAllTimeReducerState;
+  teamLeaderboardWeeklyState: TeamLeaderboardWeeklyReducerState;
+  teamLeaderboardMonthlyState: TeamLeaderboardMonthlyReducerState;
+  teamLeaderboardYearlyState: TeamLeaderboardYearlyReducerState;
+  teamLeaderboardAllTimeState: TeamLeaderboardAllTimeReducerState;
```
liammoyn, line 66: Seems like some nesting would be helpful, having a leaderboardState that then contains user and team states

src/containers/team-leaderboard/ducks/thunks.ts
```diff
@@ -0,0 +1,24 @@
...
+  LeaderboardThunkAction,
+} from '../../../components/leaderboard/ducks/types';
+import { 
+  getLeaderboardAction
+} from '../../../components/leaderboard/ducks/actions';
+
+export const getTeamsLeaderboard = (
+  previousDays: number,
+): LeaderboardThunkAction<void> => {
+  const action = getLeaderboardAction(previousDays);
```
jackblanc, line 12: We shouldn't have to have separate actions for fetching different ranges of data, we should have one action which loads the range of data as requested by the user.

liammoyn (commented):
So is our approach that we'll call the leaderboard route multiple times for each timeframe when the user loads the page? It would probably be better to only be making the call once the user actually selects a particular timeframe so that we load faster and have less data overhead

src/components/leaderboard/ducks/actions.ts
```diff
@@ -0,0 +1,39 @@
+import { genericAsyncActions } from '../../../utils/asyncRequest';
```
jackblanc, line 1: Why is this in the components file? Shouldn't it be in `user-leaderboard` and `team-leaderboard` containers?

jackblanc (changes requested):
Overall this looks incredible, thank you for doing this, and I'm sorry I didn't get you a review any sooner. 

This is mostly there, but I think there's a lot of extra code we won't end up needing. 

The main idea I have in this review is that we want to simplify the reducers state and remove usages of React state. We don't need a separate reducer for each time frame of each leaderboard. If we want to store all four time frames in state, we can make four pieces of state in each reducer. Personally, I think we're better off simplifying this by simply having one reducer per leaderboard, which can be called to fetch T - X days for any X.

Let me know if any of this is confusing, happy to chat during office hours today.

src/containers/team-leaderboard/ducks/thunks.ts
willmt80: When I had one action before, it was not making a distinction between each of the ranges, so everything got set to the all time data since that was the once dispatched last. However this won't be an issue if we change this to only store one tab's data at a time

src/components/leaderboard/ducks/actions.ts
willmt80: So that we only need one set of actions for both leaderboards instead of having 2 identical action files in each container.

willmt80:
Condensed a lot of the redux stuff so there is now only 1 item in the store for each of the leaderboards. Also moved the logic for converting LeaderboardItems to TabItems in order to avoid having a redundant React state

src/api/apiClient.ts
```diff
@@ -0,0 +1,47 @@
...
+  ) => Promise<LeaderboardItem[]>;
+}
+
+enum ApiClientRoutes {
+  USERS_LEADERBOARD = '/api/v1/leaderboard/users',
+  TEAMS_LEADERBOARD = '/api/v1/leaderboard/teams',
+}
+
+const getUsersLeaderboard = (
+  previousDays: number,
```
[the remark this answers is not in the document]
willmt80: It's not required, the backend sets a default if nothing is passed. Right now that default is 100. As it stands idk if there is a way to truly get "all time" without just passing a very large number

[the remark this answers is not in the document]
jackblanc: If it's not required I would make it optional.

Also please change this back to actually hit the API for final review.

src/components/leaderboard/ducks/actions.ts
jackblanc: I think it makes more sense to have two different action files, that way if they diverge then logically we don't have to worry about rewriting.

src/components/leaderboard/ducks/types.ts
```diff
@@ -0,0 +1,17 @@
...
+import { LeaderboardItemAction } from './actions';
+import { ApiExtraArgs } from '../../../api/apiClient';
+
+export interface LeaderboardItem {
+  id: number;
+  name: string;
+  blocksCounted: number;
+}
+
+export type LeaderboardThunkAction<R> = ThunkAction<
```
jackblanc, line 12: I think this should also be in a `types` file in each of the ducks folders as opposed to one location

src/components/leaderboard/leaderboard-tab/LeaderboardTab.tsx
```diff
@@ -48,13 +49,26 @@ interface LeaderboardTabProps {
...
-export interface LeaderboardItem {
+export interface TabItem {
   rank?: number;
+  id: number;
   name: string; // for now, these are assumed to be unique
   rightSide: React.ReactNode;
   collapseContent?: React.ReactNode;
 }
 
+export const leaderboardItemsToTabItems = (
```
jackblanc, line 60: Should this function be exported? I don't think it should be used outside of this component.

src/components/linkcarousel/LinkCarousel.tsx
```diff
@@ -58,7 +58,11 @@ const LinkCarousel: React.FC = () => {
           />
         </CarouselSlide>
         <CarouselSlide>
-          <LinkCard text="Team Leaderboard" path="/" background="img4" />
+          <LinkCard
+            text="Team Leaderboard"
+            path="/team-leaderboard"
+            background="img4"
```
jackblanc, line 64: what is img4? Where is it stored?

src/containers/team-leaderboard/TeamLeaderboard.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+const TeamLeaderboard: React.FC<TeamLeaderboardProps> = ({
+  teamLeaderboardItems,
+}) => {
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    dispatch(getTeamsLeaderboard(LEADERBOARD_DAYS.weekly));
+  }, []);
```
jackblanc, line 36: Does dispatch need to be wrapped in a useEffect? Honestly not sure if it does 🤔

src/containers/volunteer-leaderboard/VolunteerLeaderboard.tsx
```diff
@@ -53,19 +22,52 @@ const LeaderboardContainer = styled.div`
...
+  readonly userLeaderboardItems: UserLeaderboardReducerState['userLeaderboard'];
+}
+
+const VolunteerLeaderboard: React.FC<VolunteerLeaderboardProps> = ({
+  userLeaderboardItems,
+}) => {
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    // default
```
jackblanc, line 35:
```suggestion
```

Either write a more complete comment or remove this

src/store.ts
```diff
@@ -15,31 +15,48 @@ import userReducer, { initialUserState } from './auth/ducks/reducers';
...
-export type C4CAction = UserAuthenticationActions;
+export type C4CAction = UserAuthenticationActions & LeaderboardItemAction;
 
-export type ThunkExtraArgs = UserAuthenticationExtraArgs;
+export type ThunkExtraArgs = UserAuthenticationExtraArgs & ApiExtraArgs;
 
 const reducers = combineReducers<C4CState, C4CAction>({
   authenticationState: userReducer,
+  userLeaderboardState: userLeaderboardReducer,
+  teamLeaderboardState: teamLeaderboardReducer,
```
jackblanc, line 47: 👍

jackblanc (commented):
This looks much better! Left a few more questions, please update the API client to hit the API and verify that it can connect properly, then I'll give a final review.

src/components/linkcarousel/LinkCarousel.tsx
willmt80: In the linkcard, folder. I didn't make this component, I just added the path variable and prettier changed the indenting

jackblanc:
Screen Shot 2021-02-15 at 1 16 38 PM
The issues with checks are valid and must be fixed.

src/containers/team-leaderboard/TeamLeaderboard.tsx
willmt80: Yeah it's necessary, get a maximum call stack exceeded error without it

src/auth/axios.ts
```diff
@@ -19,6 +18,7 @@ const listener = () => {
 };
 
 store.subscribe(listener);
```
willmt80, line 20: Is this used anywhere? I got an error when it wasn't commented out

jackblanc: Hmm, it is, I'll look into this right now it came up yesterday for me too

jackblanc (approved):
Looks great! Awesome work iterating on this

jackblanc:
@willmt80 let me know if you want help resolving these merge conflicts

PR #30, opened 2021-02-01
Sd.mobile landing page

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/925485550)

Allows the user to navigate the landing page of the site on a mobile device more easily. Creates components that can be reused when creating other mobile pages.

## This PR

Global Styling
- renamed `DARKGREY` to `DARK_GREY` for consistency
- added `BACKGROUND_GREY`

Footer for Mobile
- created `MobileFooter` to match footer in mobile designs

Nav Bar
- created a separate component for `MobileNavBar` (separate component to improve readability)
- `NavBar` determines if the device is mobile and whether the user is logged in, renders `MobileNavBar` if mobile and passes `isLoggedIn`
- mobile nav bar has a dropdown menu if user is logged in

Create map page for mobile
- created `MobileMapPage` component
- compared to `MapPage`, children are displayed below the map rather than as a sidebar
- scroll down to see `MobileFooter`

Optimize map view for mobile
- `MapView` does not render address search bar if mobile

Create landing bar for mobile
- created `MobileLandingBar` component to match designs

Optimize landing page for mobile
- renders `MobileMapPage` if mobile

Create info card for mobile
- created `MobileInfoCard` component to match landing stats card in mobile designs (rather than optimizing `InfoCard` to reduce use of `useWindowDimensions`)

Optimize landing stats for mobile, use styled components
- deleted `landing-tree-stats.less` and used styled components
- displays stats in `MobileInfoCard` if mobile, otherwise in `InfoCard`

## Screenshots

 
Mobile Landing Page with Mobile Nav Bar (left), Mobile Landing Page with Mobile Footer (right)
MobileLanding  MobileFooter
Dropdown Menu on Mobile Nav Bar (displayed when logged in)
MobileNavDropdown

## Verification Steps

 

Used Device Mode in Chrome Dev Tools to check the mobile layout and map functionality, clicked the dropdown menu and links to make sure they worked.


src/components/landing-tree-stats/LandingTreeStats.tsx
```diff
@@ -14,33 +50,76 @@ const LandingTreeStats: React.FC<LandingTreeStatsProps> = ({
...
+          </MobileMapCardsContainer>
+
+          <GreyParagraph>
+            Learn more about how we got these numbers{' '}
+            <Link underline>here</Link>.
+          </GreyParagraph>
+        </TreeStatsContainer>
+      );
+
+    default:
```
jackblanc, line 88: Rather than relying upon the default case, you should explicitly show the desktop case as well. A potential problem this could bring is that if you add a new type to the enum, it will assume this default behavior, even if that is not logically wanted.

src/components/map-page-components/map-view/MapView.tsx
```diff
@@ -275,7 +278,9 @@ const MapView: React.FC = () => {
   return (
     <>
       <div id="pac-container">
-        <StyledSearch id="pac-input" placeholder="Address" />
+        {!(windowType === WindowTypes.Mobile) && (
```
jackblanc, line 281:
```suggestion
        {windowType !== WindowTypes.Mobile && (
```
This? Or, `windowType === WindowTypes.Desktop`?

src/components/map-page-components/mobile-landing-bar/MobileLandingBar.tsx
```diff
@@ -0,0 +1,114 @@
+import React from 'react';
```
jackblanc, line 1: I probably should have uh written this down somewhere, or changed frontend scaffold, idk, but we should be using `lowerCamelCase` for directory names. 

You can also name the component `index.tsx` and import it without specifying the full file path. 

No changes required on these, but if you want to :)

src/components/map-page-components/mobile-landing-bar/MobileLandingBar.tsx
```diff
@@ -0,0 +1,114 @@
...
+    <>
+      <MobileBarContentContainer>
+        <TitleButtonsContainer>
+          <TitleContainer>
+            <MobileTitle>Boston's Street Trees</MobileTitle>
+            <MobileParagraph>
+              Dreamcatcher kogi taiyaki keytar. Swag typewriter craft beer
+              cronut pok pok gentrify flannel salvia deep v pork belly
+              pitchfork. Swag fashion axe fam. Occupy biodiesel jean shorts
+              affogato PBR&B freegan bushwick vegan four loko pickled.
```
jackblanc, line 82: Can we please get rid of this or make it somewhat more acceptable 😂 I blame Brandon Yap

src/components/map-page-components/mobile-landing-bar/MobileLandingBar.tsx
```diff
@@ -0,0 +1,114 @@
...
+            </MobileParagraph>
+          </TitleContainer>
+          <ButtonsContainer>
+            <div>
```
jackblanc, line 86: Could use `<>` instead of `<div>` (react fragment)

src/components/map-page-components/mobile-map-page/MobileMapPage.tsx
```diff
@@ -0,0 +1,30 @@
...
+const MainContent = styled.div`
+  height: 100%;
+  margin-bottom: 15px;
+`;
+
+const MobileMapPage: React.FC = ({ children }) => {
+  return (
+    <>
+      <MainContent>
+        <Layout style={{ height: 'calc(100vh - 83px)' }}>
```
jackblanc, line 18: This is a powerful string to have in an inline style. I believe this height calculation is used many places, can it be abstracted to a constant somewhere?

src/components/mobile-nav-bar/MobileNavBar.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const MobileLoggedInExtra = () => {
+    return (
+      <FlexDiv>
+        <Dropdown overlay={mobileMenu} placement="bottomLeft">
+          <MobileDropdownMenu />
+        </Dropdown>
+      </FlexDiv>
+    );
+  };
```
jackblanc, line 96: Personally, I think it'd be cleaner to have these sub-renders included inline below.

src/containers/landing/Landing.tsx
```diff
@@ -1,31 +1,66 @@
...
+  const statMoneySaved = 100000;
+  const statRainWater = 100000;
+  const statCarbonEmissions = 31;
+
+  switch (windowType) {
+    case WindowTypes.Mobile:
+      return (
+        <>
+          <Helmet>
+            <title>Landing</title>
```
jackblanc, line 23: Let's rename the title, this sets the browser tab content

src/containers/landing/Landing.tsx
```diff
@@ -1,31 +1,66 @@
...
+                moneySaved={statMoneySaved}
+                rainWater={statRainWater}
+                carbonEmissions={statCarbonEmissions}
+              />
+            </MobileLandingBar>
+          </MobileMapPage>
+        </>
+      );
+
+    default:
```
jackblanc, line 41: Same thing with explicitly showing cases here

jackblanc (commented):
This looks great! Left a few comments on room for improvement :)

src/components/landing-tree-stats/LandingTreeStats.tsx
SofieDunt: 👍 The default case is used like this in other components too where the desktop, tablet, and narrow desktop share the same styling - I can open another PR to fix that later. Since the default case is required, should we have that display an error message - something like "This browser type is not supported"  (as the types are currently defined, this would never render)?

src/components/map-page-components/mobile-landing-bar/MobileLandingBar.tsx
SofieDunt: I can open a PR to fix all the other directory names too.

src/components/map-page-components/mobile-landing-bar/MobileLandingBar.tsx
SofieDunt: This shows up quite a bit - is there actually going to be a description to replace it later?

src/components/map-page-components/mobile-landing-bar/MobileLandingBar.tsx
SofieDunt: Using `<>` changes the styling

src/components/map-page-components/mobile-map-page/MobileMapPage.tsx
SofieDunt: We could put the constant somewhere - maybe in `window-dimensions.tsx`? Alternatively, I think it's used for the map layouts only so we could make a `MapLayout` component with this inline styling.

src/containers/landing/Landing.tsx
SofieDunt: `Speak for the Trees`?

src/components/landing-tree-stats/LandingTreeStats.tsx
jackblanc: When different cases share the same functionality, you can group them together as follows: 
```
switch (widnowType)
  case WindowTypes.Mobile: 
    // code
    break;
  case WindowTypes.Desktop:
  case WindowTypes.NarrowDesktop:
    // code for both 
    break;  
```

As for what the default case should return, an error message would be good, I think it'd also be acceptable to make  it return a React Fragment, which is basically like returning null JSX `<></>`.

src/components/map-page-components/mobile-landing-bar/MobileLandingBar.tsx
jackblanc: That would be excellent, cc @florisdobber would be good to track this in monday

src/components/map-page-components/mobile-landing-bar/MobileLandingBar.tsx
jackblanc: Reine was working on writing copy but I don't think she's had time for it. @florisdobber have we asked SFTT what they want the copy to be or are they expecting us to come up with it all? Honestly seems like something they should just do.

src/components/map-page-components/mobile-landing-bar/MobileLandingBar.tsx
jackblanc: Ok cool, then it's fine 👍

src/components/map-page-components/mobile-map-page/MobileMapPage.tsx
jackblanc: If it's only map pages, then I think a MapLayout component would be good. Otherwise use window dimensions.

src/containers/landing/Landing.tsx
jackblanc: Perfect

jackblanc (commented):
Looks great, I think the main remaining thing is to update the switch cases, unless I just can't read.

src/components/map-page-components/mapLayout/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { Layout } from 'antd';
+
+const mapLayoutHeight = 'calc(100vh - 83px)';
+
+const MapLayout: React.FC = ({ children }) => {
+  return (
+    <>
+      <Layout style={{ height: mapLayoutHeight }}>{children}</Layout>
```
jackblanc, line 9: Is there a reason we're setting this inline as opposed to using styled-components?

jackblanc (approved):
Looks great! Nice work on this

PR #31, opened 2021-02-01
reservation routes in protected api client

willmt80:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/925415585)

This PR connects the reservation routes to the frontend to allow users to connect to the sftt backend and edit the status of blocks

## This PR

Added a new protected API client file and added each of the routes and functions. Followed the same pattern as the routes made by @jackblanc in LLB.

## Verification Steps

The UI for the reservation pages is not fully set up so I have not been able to confirm that everything is working as expected.


jackblanc (commented):
Nice work! Take a look at the `authClient.test.ts` for a guide on how to test this, it should be pretty simple but I think it's worth having.

You may need to include `.then.catch` chains to resolve the promises... honestly not too sure. We can cross that bridge when we use these.

willmt80:
The fail_qa route is not recognized. Not sure why

src/api/test/protectedApiClient.test.ts
```diff
@@ -0,0 +1,135 @@
+import nock from 'nock';
+import ProtectedApiClient, {
+  ProtectedApiClientRoutes,
+  AdminApiClientRoutes,
+} from '../protectedApiClient';
+
+const BASE_URL = 'http://localhost';
+
+// TODO: handle invalid cases
```
jackblanc, line 9: Can you add a ticket for this on Monday on the bugs/tech debt section? Don't want this to get lost. Or finish it as part of this PR.

jackblanc (approved):
Looks good, we love tests :)

PR #34, opened 2021-02-08
Redirect After Login

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/1016837535)

Redirects the user to the home page after they log in/sign up. Creates a `ROUTE` enum to make altering routes easier.

## This PR

- Get privilege level to check whether the user is logged in, if they are, redirect to home
- If the login/signup request fails, displays an error message
- Edited login page and form to list all window types in the switch case
- Login form checks whether the input in the email field follows the email pattern
- Styling changes to make tablet views neater (no overflow on signup form)

## Screenshots

 
LoginError
SignupError

## Verification Steps

Ran frontend and backend locally, checked all window types:
- entered incorrect login to see if error displayed correctly
- entered correct login for a user with standard privilege level to see redirect to home
- entered correct login for a user with admin privilege level to see redirect to home 
- entered signup request with an invalid username to see if error displayed correctly
- entered signup request with an invalid email to see if error displayed correctly
- entered valid signup request to see redirect to home


src/containers/login/Login.tsx
```diff
@@ -54,17 +59,47 @@ const Footer: typeof Paragraph = styled(Paragraph)<ParagraphProps>`
...
 type LoginProps = UserAuthenticationReducerState;
 
 const Login: React.FC<LoginProps> = ({ tokens }) => {
   const { windowType, width } = useWindowDimensions();
   const dispatch = useDispatch();
-  const onFinish = (values: any) => {
+  const history = useHistory();
+  const privilegeLevel = useSelector((state: C4CState) =>
+    getPrivilegeLevel(tokens),
+  );
```
jackblanc, line 89:
```suggestion
  const privilegeLevel = useSelector((state: C4CState) =>
    getPrivilegeLevel(state.authenticationState.tokens),
  );
```

This selector should reference the state that it is provided rather than the prop. This could have effects on the Redux lifecycle/rules of hooks. Small issue though, probably fine either way.

src/containers/login/Login.tsx
```diff
@@ -54,17 +59,47 @@ const Footer: typeof Paragraph = styled(Paragraph)<ParagraphProps>`
...
+  const [loginFailed, setLoginFailed] = useState(false);
```
jackblanc, line 90: We want to avoid duplicating state and instead rely upon our global state here. 
```suggestion
```

src/containers/login/Login.tsx
```diff
@@ -54,17 +59,47 @@ const Footer: typeof Paragraph = styled(Paragraph)<ParagraphProps>`
...
+  const isLoggedIn: boolean =
+    privilegeLevel === PrivilegeLevel.STANDARD ||
+    privilegeLevel === PrivilegeLevel.ADMIN;
+
+  const onFinish = (values: LoginRequest) => {
     dispatch(login({ email: values.email, password: values.password }));
+    isLoggedIn
+      ? history.push('/home')
+      : setTimeout(() => setLoginFailed(true), 100);
   };
```
jackblanc, line 101:
```suggestion    
  if (privilegeLevel !== NONE) {
    history.push('/home')
  }

  const onFinish = (values: LoginRequest) => {
    dispatch(login({ email: values.email, password: values.password }));
  };
```

Essentially, this component will always be checking the global state to determine if the user is signed in. Whenever the user's privilege level is not NONE, we should immediately redirect them away from login.

It may seem odd to be checking privilegeLevel before we make the request, to be doing those independently, but this is because we are relying upon global state.

jackblanc: Another side note: would be incredible if you could make a routes enum for `/home` and replace occurrences of each, that way we can alter routes with greater ease.

src/components/nav-bar/NavBar.tsx
```diff
@@ -1,23 +1,30 @@
...
 
   const privilegeLevel = useSelector((state: C4CState) =>
-    getPrivilegeLevel(state.authenticationState.tokens),
+    getPrivilegeLevel(tokens),
   );
 
-  const isLoggedIn: boolean = privilegeLevel > PrivilegeLevel.NONE;
+  const isLoggedIn: boolean =
+    privilegeLevel === PrivilegeLevel.STANDARD ||
+    privilegeLevel === PrivilegeLevel.ADMIN;
```
jackblanc, line 27:
```suggestion
  const isLoggedIn: boolean = privilegeLevel !== PrivilegeLevel.NONE
```

jackblanc (commented):
This is 99% there, looks great, just want to avoid the use of local state in the login container. 

Could you also add this change for the Signup page while you're at it?

jackblanc (commented):
```
{tokens.kind === AsyncRequestKinds.Failed && (
    <Paragraph>{tokens.error}</Paragraph>
)}
```
You have `token` in scope for the Login container, which is of type `AsyncRequest<TokenPayload, any>`. In this case `TokenPayload` is the response type and `any` is the error type. Each `AsyncRequest` has a `kind` property, which is one of NotStarted, Loading, Loaded, or Failed. Loading and Failed are unique in that they grant you access to the types you specified. Currently, the error type is any, but it's a string response. So, if the request is Failed you can access the `error` field. You don't have to use JSX like in the example, you can do anything with the value.

src/App.tsx
```diff
@@ -31,6 +31,17 @@ const { Content } = Layout;
 
 type AppProps = UserAuthenticationReducerState;
 
+export enum ROUTE {
```
jackblanc, line 34: 🔥

jackblanc, line 34: For consistency, I named this `Routes` on the frontend. Enum values are uppercase, so that's correct. Could you update this to match?

src/components/nav-bar/NavBar.tsx
```diff
@@ -32,15 +36,17 @@ const FlexDiv = styled.div`
...
+const NavBar: React.FC<NavBarProps> = ({ tokens }) => {
   const { windowType } = useWindowDimensions();
   const history = useHistory();
 
   const privilegeLevel = useSelector((state: C4CState) =>
     getPrivilegeLevel(state.authenticationState.tokens),
   );
 
-  const isLoggedIn: boolean = privilegeLevel > PrivilegeLevel.NONE;
+  const isLoggedIn: boolean = privilegeLevel !== PrivilegeLevel.NONE;
```
jackblanc, line 49: 👍

jackblanc (approved):
Nice work! See comment on refactoring `ROUTE` to `Routes`, you should be able to use your IDE to change all occurrences automatically.

src/App.tsx
```diff
@@ -54,68 +65,84 @@ const App: React.FC<AppProps> = ({ tokens }) => {
...
                     </Switch>
                   );
 
                 case PrivilegeLevel.STANDARD:
                   return (
                     <Switch>
-                      <Route path="/" exact component={Landing} />
-                      <Route path="/login">
-                        <Redirect to="/home" />
+                      <Route path={Routes.LANDING} exact component={Landing} />
```
willmt80, line 94: Do we want this route to redirect to the home page if someone is logged in?

PR #35, opened 2021-02-08
Authentication for Protected Pages

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/1016839076)

Ensures only users with proper privilege levels can see protected pages.

## This PR

- Updated `PrivilegeLevel` and `getPrivilegeLevel()`
- Use switch case to declare routes for each privilege level type
- If the user lacks the appropriate privilege level for a path, redirects using `Redirect`

## Verification Steps

 
Ran frontend and backend locally, tried each path and checked it went to the appropriate place after:
- not logging in
- logging in as a standard user
- logging in as an admin

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -32,7 +32,7 @@ const { Content } = Layout;
 type AppProps = UserAuthenticationReducerState;
 
 const App: React.FC<AppProps> = ({ tokens }) => {
-  const privilegeLevel = useSelector((state: C4CState) =>
+  const privilegeLevel: PrivilegeLevel = useSelector((state: C4CState) =>
     getPrivilegeLevel(tokens),
   );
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -14,9 +26,16 @@ import VolunteerLeaderboard from './containers/volunteer-leaderboard/VolunteerLe
...
 import NavBar from './components/nav-bar/NavBar';
 import { Layout } from 'antd';
+
 const { Content } = Layout;
 
-const App: React.FC = () => {
+type AppProps = UserAuthenticationReducerState;
+
+const App: React.FC<AppProps> = ({ tokens }) => {
+  const privilegeLevel = useSelector((state: C4CState) =>
```
jackblanc, line 35:
```suggestion
  const privilegeLevel: PrivilegeLevel = useSelector((state: C4CState) =>
```

src/App.tsx
```diff
@@ -26,26 +45,98 @@ const App: React.FC = () => {
...
-              <Route path="/settings" exact component={Settings} />
-              <Route path="/volunteer" exact component={VolunteerLeaderboard} />
-              <Route path="/admin" exact component={AdminDashboard} />
-              <Route path="*" exact component={NotFound} />
-            </Switch>
+            {(() => {
+              switch (privilegeLevel) {
+                case PrivilegeLevel.NONE:
+                  return (
+                    <>
```
jackblanc, line 56: I don't think these fragments are required

src/App.tsx
```diff
@@ -26,26 +45,98 @@ const App: React.FC = () => {
...
+                          exact
+                          component={VolunteerLeaderboard}
+                        />
+                        <Route path="/admin" exact component={AdminDashboard} />
+                        <Route path="*" exact component={NotFound} />
+                      </Switch>
+                    </>
+                  );
+              }
+            })()}
```
jackblanc, line 128: Nice use of the immediately envoked function expression 👍

src/App.tsx
```diff
@@ -26,26 +45,98 @@ const App: React.FC = () => {
...
           </Content>
         </Layout>
       </Router>
     </>
   );
 };
 
-export default App;
+const mapStateToProps = (state: C4CState): AppProps => {
```
jackblanc, line 136: 👍

src/auth/ducks/types.ts
```diff
@@ -46,8 +46,8 @@ export interface RefreshTokenResponse {
 
 export enum PrivilegeLevel {
   NONE = -1,
-  STANDARD = 0,
-  ADMIN = 1,
+  STANDARD = 'standard',
```
jackblanc, line 49: Is the backend all uppercase? should this match?

src/auth/token.tsx
```diff
@@ -31,7 +31,7 @@ export interface TokenService {
   readonly getRefreshToken: () => string | null;
   readonly setRefreshToken: (token: string) => void;
   readonly removeRefreshToken: () => void;
-  readonly getPrivilegeLevel: () => number;
+  readonly getPrivilegeLevel: () => number | string;
```
jackblanc, line 34: We should keep this as all number or all string, the enum above can be updated so that NONE = 'none'.

src/App.tsx
```diff
@@ -26,26 +45,98 @@ const App: React.FC = () => {
...
+                case PrivilegeLevel.NONE:
+                  return (
+                    <>
+                      <Switch>
+                        <Route path="/" exact component={Landing} />
+                        <Route path="/login" exact component={Login} />
+                        <Route path="/signup" exact component={Signup} />
+                        <Route path="/home">
+                          <Redirect to="/login" />
+                        </Route>
```
jackblanc, line 63: Didn't know you could do this, nice

jackblanc (approved):
Nice work :)

src/auth/ducks/types.ts
SofieDunt: Not too sure but all uppercase doesn't work here (default in backend seems to be lowercase `standard`)

src/auth/token.tsx
SofieDunt: Updated `getPrivilegeLevel()` too

PR #37, opened 2021-02-08
Add Interface to Window Dimensions

jackblanc:
## Why

We are missing an interface on the window dimension utility functions. 

## This PR

Adds an interface for window dimensions. 

## Screenshots

No visual changes. 

 

## Verification Steps

Code builds.


florisdobber (approved):
Looks good to me! Gets rid of one more error

PR #41, opened 2021-02-22
Settings routes

willmt80:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/925294524)

This PR connects the remaining user settings routes to the frontend. This will allow users to delete their accounts, change usernames, and it will allow admins to change the privilege levels of other users

## This PR

Added new routes to the protected api routes file, created a new test file and directory, and added to the settings and admin forms pages.

## Screenshots

 

No new styling

## Verification Steps

 

I created tests for the cases where the routes work as planned. I could not figure out how to handle errors so that will be an iteration onto this PR. I also tested the change_privilege form with one the the users in my local backend.

### DIFF
--- a/src/api/protectedApiClient.ts
+++ b/src/api/protectedApiClient.ts
@@ -5,10 +5,14 @@ export interface ProtectedApiClient {
     currentPassword: string;
     newPassword: string;
   }) => Promise<void>;
-  /*readonly changeUsername: (request: {
+  readonly changeUsername: (request: {
     newUsername: string;
     password: string;
-  }) => Promise<void>; */
+  }) => Promise<void>;
+  readonly changeEmail: (request: {
+    newEmail: string;
+    password: string;
+  }) => Promise<void>;
   readonly deleteUser: (request: { password: string }) => Promise<void>;
   readonly changePrivilegeLevel: (request: {
     targetUserEmail: string;
@@ -19,7 +23,8 @@ export interface ProtectedApiClient {
 
 export enum ProtectedApiClientRoutes {
   CHANGE_PASSWORD = '/api/v1/protected/user/change_password',
-  // CHANGE_USERNAME = '/api/v1/protected/user/change_username',
+  CHANGE_USERNAME = '/api/v1/protected/user/change_username',
+  CHANGE_EMAIL = '/api/v1/protected/user/change_email',
   DELETE_USER = '/api/v1/protected/user/',
 }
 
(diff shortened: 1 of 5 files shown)


src/api/protectedApiClient.ts
```diff
@@ -5,10 +5,26 @@ export interface ProtectedApiClient {
     currentPassword: string;
     newPassword: string;
   }) => Promise<void>;
+  /*readonly changeUsername: (request: {
```
jackblanc, line 8: Does this not work or exist?

src/containers/adminDashboard/index.tsx
```diff
@@ -17,26 +19,27 @@ const EditUser = styled.div`
...
-
 const AdminDashboard: React.FC = () => {
   const isSuperAdmin = true;
 
-  const roleList: DropdownItem[] = [
-    { key: 1, label: 'Standard user' },
-    { key: 2, label: 'Admin' },
-    { key: 3, label: 'Super admin' },
-  ];
+  const formRef = useRef<FormInstance>(null);
```
jackblanc, line 30: Why this over useState?

jackblanc (approved):
This looks great! If you could apply these changes to frontend-scaffold as well that would be incredible. Also not sure what's up with changeUsername, it's commented out.

src/containers/adminDashboard/index.tsx
willmt80: The other forms on the settings pages use form submission instead of having a callback that changes the state. I don't think it makes a difference functionally

willmt80:
The delete users route does not work. The backend uses a .delete request which does not accept a request body

src/containers/settings/index.tsx
```diff
@@ -68,15 +90,33 @@ const Settings: React.FC = () => {
...
+              <Form.Item
+                {...formLayout}
+                name="new-email"
+                rules={[
+                  {
+                    required: true,
+                    message: 'Please input a new email address!',
+                  },
+                  {
+                    pattern: /^\S+@\S+\.\S{2,}$/,
```
jackblanc, line 115: This pattern should probably be extracted somewhere/compared against whatever we use on the backend

jackblanc (commented):
Looks like this just has to change to a POST request, awesome work on the test coverage

src/containers/settings/index.tsx
florisdobber: Checked and matches backend

PR #42, opened 2021-02-28
Individual Teams Page

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/926124580)

Creates the team page where users can view a team's members and goals and apply to join.

## This PR

- Created a `types.ts` to match backend
- Created `TeamPage` container, which uses a list to display all members in the team and a collapse to display goals
  - info about members (username, role) is formatted in `TeamMember` component
    - team leader has a crown next to their username
  - expanding a collapse panel shows info about the goal, which is formatted in `GoalInfo` component
  - the first goal in the data is expanded by default
  - "Incomplete" is displayed by the collapse panel's dropdown if the goal's `completion_date` is `null`
- Added route for individual team page
- Created `getDateString` to format dates as a string, moved it to `utils/stringFormat.tsx` with `getMoneyString` (and tests)
- "Return to Teams" button currently redirects to login page, should be changed after teams page is added

## Screenshots

 
IndividualTeamPage
OpenAllGoals

## Verification Steps

 
Used dummy data, visually confirmed data was displayed correctly and page matched designs

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -75,12 +75,26 @@ const App: React.FC = () => {
                       <Route path={Routes.LANDING} exact component={Landing} />
                       <Route path={Routes.LOGIN} exact component={Login} />
                       <Route path={Routes.SIGNUP} exact component={Signup} />
+                      <Route path={Routes.HOME}>
+                        <Redirect to={Routes.LOGIN} />
+                      </Route>
+                      <Route path={Routes.SETTINGS}>
+                        <Redirect to={Routes.LOGIN} />
+                      </Route>
+                      <Route path={Routes.VOLUNTEER}>
+                        <Redirect to={Routes.LOGIN} />
+                      </Route>
+                      <Route path={Routes.TEAM_LEADERBOARD}>
+                        <Redirect to={Routes.LOGIN} />
+                      </Route>
+                      <Route path={Routes.ADMIN}>
+                        <Redirect to={Routes.LOGIN} />
+                      </Route>
                       <Route
                         path={Routes.NOT_FOUND}
                         exact
                         component={NotFound}
                       />
-                      <Redirect to={Routes.LOGIN} />
                     </Switch>
                   );
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -43,7 +43,7 @@ export enum Routes {
   HOME = '/home',
   SETTINGS = '/settings',
   VOLUNTEER = '/volunteer',
-  TEAM = '/team',
+  TEAM = '/team:id',
   TEAM_LEADERBOARD = '/team-leaderboard',
   ADMIN = '/admin',
   NOT_FOUND = '*',
@@ -75,29 +75,12 @@ const App: React.FC = () => {
                       <Route path={Routes.LANDING} exact component={Landing} />
                       <Route path={Routes.LOGIN} exact component={Login} />
                       <Route path={Routes.SIGNUP} exact component={Signup} />
-                      <Route path={Routes.HOME}>
-                        <Redirect to={Routes.LOGIN} />
-                      </Route>
-                      <Route path={Routes.SETTINGS}>
-                        <Redirect to={Routes.LOGIN} />
-                      </Route>
-                      <Route path={Routes.VOLUNTEER}>
-                        <Redirect to={Routes.LOGIN} />
-                      </Route>
-                      <Route path={Routes.TEAM}>
-                        <Redirect to={Routes.LOGIN} />
-                      </Route>
-                      <Route path={Routes.TEAM_LEADERBOARD}>
-                        <Redirect to={Routes.LOGIN} />
-                      </Route>
-                      <Route path={Routes.ADMIN}>
-                        <Redirect to={Routes.LOGIN} />
-                      </Route>
                       <Route
                         path={Routes.NOT_FOUND}
                         exact
                         component={NotFound}
                       />
+                      <Redirect to={Routes.LOGIN} />
                     </Switch>
                   );
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,6 +19,7 @@ import Landing from './containers/landing';
 import AdminDashboard from './containers/adminDashboard';
 import VolunteerLeaderboard from './containers/volunteerLeaderboard';
 import TeamLeaderboard from './containers/teamLeaderboard';
+import TeamPage from './containers/teamPage';
 import { Layout } from 'antd';
 import Home from './containers/home';
 import Signup from './containers/signup';
@@ -42,7 +43,8 @@ export enum Routes {
   HOME = '/home',
   SETTINGS = '/settings',
   VOLUNTEER = '/volunteer',
-  TEAM_LEADERBOARD = './team-leaderboard',
+  TEAM = '/team',
+  TEAM_LEADERBOARD = '/team-leaderboard',
   ADMIN = '/admin',
   NOT_FOUND = '*',
 }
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -42,6 +43,7 @@ export enum Routes {
   HOME = '/home',
   SETTINGS = '/settings',
   VOLUNTEER = '/volunteer',
+  TEAM = '/team',
```
willmt80, line 46: We'll want to set up the route so that it has a team ID parameter

jackblanc:
```suggestion
  TEAM = '/team/:id',
```
Agreed, this is the proper syntax I believe

src/App.tsx
```diff
@@ -82,6 +84,9 @@ const App: React.FC = () => {
                       <Route path={Routes.VOLUNTEER}>
                         <Redirect to={Routes.LOGIN} />
                       </Route>
+                      <Route path={Routes.TEAM}>
+                        <Redirect to={Routes.LOGIN} />
+                      </Route>
```
jackblanc, line 89: 📌(out of scope for this PR, something to be done on scaffold) How can we create a catch-all which simply redirects the user to log in if they try to access a page other than Landing, Login, or Signup?

src/components/landingTreeStats/index.tsx
```diff
@@ -5,6 +5,7 @@ import useWindowDimensions, { WindowTypes } from '../window-dimensions';
 import { TEXT_GREY, MID_GREEN } from '../../utils/colors';
 import MobileInfoCard from '../mobileComponents/mobileInfoCard';
 import InfoCard from '../infoCard';
+import { getMoneyString } from '../../utils/stringFormat';
```
jackblanc, line 8: 💯

src/components/teamMember/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { TeamRole } from '../../containers/teamPage/ducks/types';
+
+const { Paragraph } = Typography;
+
+const StyledListItem = styled(List.Item)`
+  margin-bottom: 10px;
+  width: 85%;
+  height: 40px;
+  background: ${LIGHT_GREEN}90;
+  border: solid 1px ${LIGHT_GREEN}90;
```
jackblanc, line 15: Is the `90` at the end intentional?

src/containers/teamPage/index.tsx
```diff
@@ -0,0 +1,268 @@
...
+};
+
+const TeamPage: React.FC = () => {
+  const onClick = () => {
+    // TODO: join team
+  };
+
+  return (
+    <TeamContainer>
+      <StyledLinkButton to={Routes.LOGIN}>
```
jackblanc, line 193: 📌  This should be a different route, no? To the "All Teams" page? Can you drop a ticket to track the work to change this?

src/containers/teamPage/index.tsx
```diff
@@ -0,0 +1,268 @@
...
+          <SectionHeader>TEAM MEMBERS</SectionHeader>
+          <StyledMemberList
+            dataSource={sampleTeamData.members}
+            itemLayout="vertical"
+          >
+            {sampleTeamData.members.map((member) => {
+              if (
+                member.team_role === TeamRole.LEADER ||
+                member.team_role === TeamRole.MEMBER
+              ) {
```
jackblanc, line 219: ❓ Why is this condition required?

src/containers/teamPage/index.tsx
```diff
@@ -0,0 +1,268 @@
...
+              return <></>;
+            })}
+          </StyledMemberList>
+        </MemberContainer>
+        <GoalContainer>
+          <SectionHeader>TEAM GOALS</SectionHeader>
+          <LeaderboardCollapse
+            bordered={false}
+            key={1}
+            defaultActiveKey={sampleTeamData.goals[0].id}
```
jackblanc, line 237: ❓ Confused what this is doing. Is key a prop we have to store in React State and control, or does the component manage that internally? If it's managed internally, why the `key={1}`?

src/containers/teamPage/index.tsx
```diff
@@ -0,0 +1,268 @@
...
+            expandIconPosition={'right'}
+          >
+            {sampleTeamData.goals.map((item) => {
+              return (
+                <StyledPanel
+                  key={item.id}
+                  header={<BlackText>Goal {item.id}</BlackText>}
+                  extra={
+                    item.completion_date == null && (
+                      <BlackText>Incomplete</BlackText>
```
jackblanc, line 247: ❓ What if the goal is complete?

src/containers/teamPage/index.tsx
```diff
@@ -0,0 +1,268 @@
...
+
+const Line = styled.div`
+  height: 1px;
+  width: 104%;
+  margin: 0px -10px 10px -20px;
+  display: block;
+  background-color: ${WHITE};
+`;
+
+// dummy data
```
jackblanc, line 107: 📌 Where is the work to connect this to the API ticketed? I can pair on implementing a Redux store for this data.

src/utils/stringFormat.tsx
```diff
@@ -0,0 +1,15 @@
+/**
```
jackblanc, line 1: Nicely done 👍

jackblanc (commented):
Looks good! Left some questions, take a look at the emoji guide if any are unclear 😂 https://github.com/erikthedeveloper/code-review-emoji-guide

src/App.tsx
SofieDunt: 👍

src/App.tsx
SofieDunt: Good point, since everything else redirects to Login, we could just only list the Landing, Login, and Signup routes (probably would want the Not Found one too) and then have a `<Redirect to={Routes.Login} />` after those four cases.

src/components/teamMember/index.tsx
SofieDunt: It sets the transparency

src/containers/teamPage/index.tsx
SofieDunt: The `TeamRole` enum also has types `None` (applied and was rejected or left the team) and `Pending` (has applied and not been accepted/rejected) - this is so we don't show those users

src/containers/teamPage/index.tsx
SofieDunt: Forgot to delete the `key={1}` - was a temporary workaround 😂 . `defaultActiveKey` expands the first goal so you can see the stats upon opening the page (instead of all panels being closed)

src/containers/teamPage/index.tsx
SofieDunt: The goal is still listed but doesn't say 'Incomplete' next to the dropdown (in the screenshots, the bottom goal is complete)

src/containers/teamPage/index.tsx
SofieDunt: [Monday](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/926123296)

src/App.tsx
SofieDunt: This was a pretty quick change so I tested and added it while fixing the team ID parameter - can undo this easily too

jackblanc: Agreed, that's a tradeoff for sure...  let's discuss next meeting

jackblanc (approved):

PR #43, opened 2021-02-28
created available teams page

willmt80:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/926124762)

This PR allows users to view which teams are accepting new people and allows the user to click a button to view the pages of those teams

## This PR

Created an `AvailableTeams` component which has a header, pagination, and the collapse with the teams

Moved a part of `LeaderboardTabs` to its own component, `LeaderboardSpace` as the main section of the page in order to reuse some of the logic for a paginated leaderboard. Added an optional `large` prop which is a boolean that sets the size of the leaderboard items.

There is an empty button on each team item right now which will eventually be a link to the individual team page

## Screenshots

 

Capture

## Verification Steps

 

Used dummy data to verify that the styling was correct. Used dummy data for a volunteer leaderboard to ensure there were no breaking changes.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,6 +19,7 @@ import Landing from './containers/landing';
 import AdminDashboard from './containers/adminDashboard';
 import VolunteerLeaderboard from './containers/volunteerLeaderboard';
 import TeamLeaderboard from './containers/teamLeaderboard';
+import AvailableTeams from './containers/availableTeams';
 import { Layout } from 'antd';
 import Home from './containers/home';
 import Signup from './containers/signup';
@@ -44,6 +45,7 @@ export enum Routes {
   VOLUNTEER = '/volunteer',
   TEAM_LEADERBOARD = '/team-leaderboard',
   ADMIN = '/admin',
+  AVAILABLE_TEAMS = '/available',
   NOT_FOUND = '*',
 }
 
@@ -85,6 +87,9 @@ const App: React.FC = () => {
                       <Route path={Routes.TEAM_LEADERBOARD}>
                         <Redirect to={Routes.LOGIN} />
                       </Route>
+                      <Route path={Routes.AVAILABLE_TEAMS}>
+                        <Redirect to={Routes.LOGIN} />
+                      </Route>
                       <Route path={Routes.ADMIN}>
                         <Redirect to={Routes.LOGIN} />
                       </Route>
@@ -122,6 +127,11 @@ const App: React.FC = () => {
                         exact
                         component={TeamLeaderboard}
                       />
+                      <Route
+                        path={Routes.AVAILABLE_TEAMS}
+                        exact
+                        component={AvailableTeams}
+                      />
                       <Route path={Routes.ADMIN}>
                         <Redirect to={Routes.HOME} />
                       </Route>
@@ -158,6 +168,11 @@ const App: React.FC = () => {
                         path={Routes.TEAM_LEADERBOARD}
                         exact
                         component={TeamLeaderboard}
+                      />
+                      <Route
+                        path={Routes.AVAILABLE_TEAMS}
+                        exact
+                        component={AvailableTeams}
                       />
                       <Route
                         path={Routes.ADMIN}
--- a/src/components/leaderboard/leaderboardSpace/index.tsx
+++ b/src/components/leaderboard/leaderboardSpace/index.tsx
@@ -1,0 +1,105 @@
+import * as React from 'react';
+import { ParagraphProps } from 'antd/lib/typography/Paragraph';
+import { SpaceProps } from 'antd/lib/space/index';
+import { BLACK, LIGHT_GREEN } from '../../../utils/colors';
+import styled from 'styled-components';
+import { Collapse, Space, Typography } from 'antd';
+
+const { Panel } = Collapse;
+const { Paragraph } = Typography;
+
+interface LeaderboardStyleProps {
+  large?: boolean;
+}
+
+const LeaderboardSpace = styled(Space)<SpaceProps>`
+  width: 100%;
+`;
+
+const LeaderboardCollapse = styled(Collapse)`
+  background-color: ${LIGHT_GREEN}80;
+  border-radius: 0px;
+  ${(props: LeaderboardStyleProps) => props.large && 'padding: 15px 0px 10px'};
+`;
+
+const LeaderboardItemName = styled(Paragraph)`
+  color: ${BLACK};
+  display: inline-block;
+  line-height: 0px;
+  ${(props: LeaderboardStyleProps) => props.large && 'font-size: 20px'};
+`;
+
+const LeaderboardRankContainer = styled.span`
+  width: 50px;
+  display: inline-block;
+`;
+
+const LeaderboardItemRank = styled(Paragraph)<ParagraphProps>`
+  color: ${BLACK};
+  display: inline;
+  vertical-align: middle;
+  text-align: center;
+  font-size: 20px;
+  line-height: 0px;
+`;
+
+export interface TabItem {
+  rank?: number;
+  id: number;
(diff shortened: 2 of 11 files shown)


src/App.tsx
```diff
@@ -85,6 +87,9 @@ const App: React.FC = () => {
                       <Route path={Routes.TEAM_LEADERBOARD}>
                         <Redirect to={Routes.LOGIN} />
                       </Route>
+                      <Route path={Routes.AVAILABLE_TEAMS}>
+                        <Redirect to={Routes.LOGIN} />
+                      </Route>
```
jackblanc, line 92: https://github.com/Code-4-Community/speak-for-the-trees-frontend/pull/42/files#r585886582

Same comment as Sofie's PR, just something to think about.

src/components/leaderboard/leaderboardSpace/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+const LeaderboardSpace = styled(Space)<SpaceProps>`
+  width: 100%;
+`;
+
+const LeaderboardCollapse = styled(Collapse)`
+  background-color: ${LIGHT_GREEN}80;
+  border-radius: 0px;
+  ${(props: LeaderboardStyleProps) => props.large && 'padding: 15px 0px 10px'};
```
jackblanc, line 22:
```suggestion
  ${({ large }: LeaderboardStyleProps) => large && 'padding: 15px 0px 10px'};
```
⛏️ Not sure if my syntax is 100% right but generally prefer to restructure the props.

src/components/leaderboard/leaderboardSpace/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+  vertical-align: middle;
+  text-align: center;
+  font-size: 20px;
+  line-height: 0px;
+`;
+
+export interface TabItem {
+  rank?: number;
+  id: number;
+  name: string; // for now, these are assumed to be unique
```
jackblanc, line 49: Will this fact change? Is the assumption valid? How is that assumption used - we have id, why do we need to assume name is also unique?

src/components/leaderboard/leaderboardSpace/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+const LeaderboardItemRank = styled(Paragraph)<ParagraphProps>`
+  color: ${BLACK};
+  display: inline;
+  vertical-align: middle;
+  text-align: center;
+  font-size: 20px;
+  line-height: 0px;
+`;
+
+export interface TabItem {
```
jackblanc, line 46: Is this duplicated from somewhere?

src/components/leaderboard/leaderboardSpace/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+  large?: boolean;
+}
+
+const LeaderboardPanels: React.FC<LeaderboardSpaceProps> = ({
+  tabItems,
+  currentPage,
+  itemsPerPage,
+  large,
+}) => {
+  const itemsOnPage = tabItems.slice(
```
jackblanc, line 68: 🔧 Add a type here for readability

src/components/leaderboard/leaderboardSpace/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+    (currentPage - 1) * itemsPerPage,
+    currentPage * itemsPerPage,
+  );
+
+  return (
+    <LeaderboardSpace direction="vertical" size={large ? 22 : 'small'}>
```
jackblanc, line 74: `size={large ? 22 : 'small'`

How does this work? 22 doesn't have a unit of measurement which is confusing.

src/components/leaderboard/leaderboardSpace/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+      {itemsOnPage.map((item) => {
+        return (
```
jackblanc, line 76:
```suggestion
      {itemsOnPage.map((item) => (
```
You can simplify this

src/components/leaderboard/leaderboardSpace/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+                    </LeaderboardItemRank>
+                  </LeaderboardRankContainer>
+                  <LeaderboardItemName large={large}>
+                    {item.name}
+                  </LeaderboardItemName>
+                </>
+              }
+              extra={item.rightSide}
+              showArrow={false}
+              disabled={!item.collapseContent}
```
jackblanc, line 94: Why `!`? Shouldn't we disable the panel when collapseContent is true?

src/components/leaderboard/leaderboardTab/index.tsx
```diff
@@ -48,55 +9,21 @@ interface LeaderboardTabProps {
...
-                      disabled={!item.collapseContent}
-                    >
-                      {item.collapseContent}
-                    </Panel>
-                  </LeaderboardCollapse>
-                );
-              })}
-            </LeaderboardSpace>
-          }
+          <LeaderboardPanels
```
jackblanc, line 22: 💯

src/components/leaderboard/leaderboardTabs/index.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { Tabs, Pagination } from 'antd';
-import LeaderboardTab, { TabItem } from '../leaderboardTab';
+import LeaderboardTab from '../leaderboardTab';
+import { TabItem } from '../leaderboardSpace';
```
jackblanc, line 4: 📌 (out of scope for this PR): it seems like many components within `components/leaderboard/` use these types. What if we stored them in a types file in `components/leaderboard/types` and referenced that instead?

jackblanc: Could be worth doing in this PR actually, but you can ticket separately if you want.

src/containers/availableTeams/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  margin-top: 10px;
+  margin-bottom: 20px;
+`;
+
+const ArrowLink = styled(ArrowRightOutlined)`
+  line-height: 0px;
+  font-size: 20px;
+`;
+
+const dummy: TabItem[] = [
```
jackblanc, line 34: Can you make sure there's a ticket for connecting this to the backend?

jackblanc (commented):
Looks great! Left some comments, let me know if there are any questions.

src/components/leaderboard/leaderboardSpace/index.tsx
willmt80: Not now, this interface was moved

src/components/leaderboard/leaderboardSpace/index.tsx
willmt80: If collapseContent is undefined, then we do not want the panel to be collapsible. Even if there is no collapseContent, unless the panel is disabled it can be opened or closed.

collapse

src/containers/availableTeams/index.tsx
willmt80: Part of the connect all team routes ticket

src/components/leaderboard/leaderboardSpace/index.tsx
willmt80: items did not originally have Ids so I just forgot to update this

src/components/leaderboard/leaderboardTabs/index.tsx
willmt80: I think it's just one type used everywhere, moved that to a new file

jackblanc (approved):

PR #44, opened 2021-03-01
Add reservation modal to map view

vrushalitarte01:
## Why

--Adds reservation modal to map so users can select blocks from map and either reserve or release the block or are notified that the block is not available.

## This PR

--I added the ReservationModal component to the MapView page and a click listener to get details on the block clicked.
--Left comments where need to complete implementation of handleOk using protectedapiclient to release a reservation
--Also need to fix how reloads map upon click
## Screenshots

 

Screen Shot 2021-03-07 at 5 18 57 PM

## Verification Steps

 

--I started the server and made sure the correct modal appears when a block is clicked on according to status


jackblanc (approved):
Sorry I forgot to approve this but it looks great :)

PR #47, opened 2021-03-13
Various frontend bug fixes

willmt80:
## Why

[Login/Signup Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1114620270)
[Vertical scroll Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1114634770)
[Header Link Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1114277426)
[Map Sidebar Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1114609771)

This PR fixes a few bugs from the sftt frontend. For one it removes vertical scrolling on desktop when text size is set to values greater than 100%.  On the landing page, the sidebar no longer extends past the length of the map and the user can now scroll through the sidebar if it is too large to fit on one screen It also adds return buttons to some pages on the frontend so that users may easily return to the home page. It also makes the text in the header logo clickable as a link to return to the home page as well.

## This PR

I tried to remove some of the explicit px settings on margins and padding because they would not always behave well on smaller screens. Instead I tried to make use of percentages of vh/vw. For the mapSidebar I added an overflow-y so that the sidebar container would remain the correct size. For the clickable header I made the text a button that when clicked takes the user to the home page. 

## Screenshots

 

landing-up

landing-down

login

sign-up

teams-with-button

## Verification Steps

 

I visited the pages both with the enlarged text (which is the default on some windows computers) and with normal 100% scaling to ensure that there was no scrolling on either.


florisdobber (commented):
I look the clean up on this! Thanks for doing all that 💯

src/App.tsx
```diff
@@ -138,6 +137,7 @@ const App: React.FC = () => {
                       <Route path={Routes.ADMIN}>
                         <Redirect to={Routes.HOME} />
                       </Route>
+                      <Route path={Routes.TEAM} exact component={TeamPage} />
```
jackblanc, line 140: It looks like the team route is incorrect, shouldn't it be /team/:id

src/containers/home/index.tsx
```diff
@@ -42,20 +37,18 @@ const Home: React.FC = () => {
...
-          subtitleColor={DARK_GREY}
-        />
-
-        <StyledSubtitle>Quick Links</StyledSubtitle>
-        <LinkCarousel />
-      </HomeContainer>
+      <PageLayout>
+        <HomeContainer>
+          <PageHeader
+            pageTitle={HOME_TITLE + userName + '!'}
```
jackblanc, line 43: ⛏️ you could use a template string: 
```
`${HOME_TITLE}${userName}!`
```

jackblanc (approved):
Nice work on this

PR #48, opened 2021-03-16
Redirect to Destination

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/1114275047)

Redirects the user to the auth page they originally wanted to go to (destination) after they log in/sign up.

## This PR

- When redirecting a user that is not logged in to the login page, passes a location object that describes the route the user wanted to go to in `location.state.destination`
  - created an `AuthRedirect` component to simplify handling this kind of redirect in `App`
  - redirects from login/signup happen within the login/signup component, not App - otherwise would redirect according to App rather than `history.push(destination)`
  - access location object in login and signup using `withRouter`, typing ensures location objects have a `location.state.destination`
- If the user went straight to `Routes.LOGIN` or `Routes.SIGNUP`, redirect goes to home
  - check if `location.state` exists, if not, set `destination` to  `Routes.HOME`

## Verification Steps

 
Tested each route when not logged in
- confirmed that: login -> home
- confirmed that: landing/signup -> login -> home
- confirmed for auth routes: destination -> login -> destination
- confirmed that: signup -> home
- confirmed that: landing/login -> signup -> home
- confirmed for auth routes: destination -> login -> signup -> destination


src/containers/login/index.tsx
```diff
@@ -90,22 +91,27 @@ const MobileLoginAlert = styled(Alert)`
   margin-bottom: 20px;
 `;
 
-type LoginProps = UserAuthenticationReducerState;
+type UserLoginProps = UserAuthenticationReducerState;
+type LoginProps = UserLoginProps & RedirectedRouteComponentProps;
 
-const Login: React.FC<LoginProps> = ({ tokens }) => {
+const Login: React.FC<LoginProps> = ({ tokens, location }) => {
```
jackblanc, line 97: Can you use `useLocation` hook instead of the higher-order-component?

src/containers/signup/index.tsx
```diff
@@ -183,10 +188,10 @@ const Signup: React.FC<SignupProps> = ({ tokens }) => {
...
 
-const mapStateToProps = (state: C4CState): SignupProps => {
+const mapStateToProps = (state: C4CState): UserSignupProps => {
   return {
     tokens: state.authenticationState.tokens,
   };
 };
 
-export default connect(mapStateToProps)(Signup);
+export default withRouter(connect(mapStateToProps)(Signup));
```
jackblanc, line 197: same as other file, you can use useLocation I think

jackblanc (approved):
Nice work, this looks like a clean, type-safe way to do this. One suggestion I have is to use the useLocation hook over the withRouter higher-order-component. Hooks typically are preferred nowadays and are newer than HOCs.

PR #49, opened 2021-03-29
Teams routes

willmt80:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1133710328)

This PR connects the team routes to the redux store and includes unit tests for these routes respond as expected. These routes will allow the user to access team data from the backend and stores it locally. It also lays the ground work for connecting that data to the frontend pages where users can view their team data as well as make requests such as joining teams.

## This PR

I added routes to the protectedApiRoutes file. Because the routes have teamIds within them, I defined most routes as functions which take in numbers and return the correct routes as strings. 

For the ducks files and testing, I followed the pattern of the previous redux tickets.

## Screenshots

 

N/A

## Verification Steps

 

I made sure each of the unit tests pass, for the get requests I used postman to verify what the objects in the store should look like.

### DIFF
--- a/src/api/protectedApiClient.ts
+++ b/src/api/protectedApiClient.ts
@@ -1,4 +1,5 @@
 import AppAxiosInstance from '../auth/axios';
+import { TeamProps, UserInvite, Applicant } from '../containers/teamPage/ducks/types';
 
 export interface ProtectedApiExtraArgs {
   readonly protectedApiClient: ProtectedApiClient;
@@ -33,6 +34,47 @@ export interface ProtectedApiClient {
     newLevel: string;
     password: string;
   }) => Promise<void>;
+  readonly createTeam: (request: {
+    name: string;
+    bio: string;
+    inviteEmails: string[];
+  }) => Promise<TeamProps>;
+  readonly getTeam: (teamId: number) => Promise<TeamProps>;
+  readonly addGoal: (
+    teamId: number,
+    request: {
+      goal: number;
+      startAt: Date;
+      completeBy: Date;
+    }
+  ) => Promise<void>;
+  readonly deleteGoal: (teamId: number, goalId: number) => Promise<void>;
+  readonly inviteUser: (
+    teamIId: number,
+    request: {
+      invites: UserInvite[]
+    }
+  ) => Promise<void>;
+  readonly getApplicants: (teamId: number) => Promise<Applicant[]>;
+  readonly applyToTeam: (teamId: number) => Promise<void>;
+  readonly approveUser: (
+    teamId: number,
+    userId: number,
+  ) => Promise<void>;
+  readonly rejectUser: (
+    teamId: number,
+    userId: number,
+  ) => Promise<void>;
+  readonly kickUser: (
+    teamId: number,
+    memberId: number,
+  ) => Promise<void>;
+  readonly leaveTeam: (teamId: number) => Promise<void>;
+  readonly disbandTeam: (teamId: number) => Promise<void>;
+  readonly transferOwnership: (
+    teamId: number,
+    request: { newLeaderId: number }
+  ) => Promise<void>;
 }
 
 export enum ProtectedApiClientRoutes {
@@ -43,6 +85,7 @@ export enum ProtectedApiClientRoutes {
   CHANGE_USERNAME = '/api/v1/protected/user/change_username',
   CHANGE_EMAIL = '/api/v1/protected/user/change_email',
   DELETE_USER = '/api/v1/protected/user/',
+  CREATE_TEAM = '/api/v1/protected/teams/create',
 }
 
 export enum AdminApiClientRoutes {
@@ -52,6 +95,8 @@ export enum AdminApiClientRoutes {
   FAIL_RESERVATION_QA = '/api/v1/protected/reservations/fail_qa',
   CHANGE_PRIVILEGE = '/api/v1/protected/user/change_privilege',
 }
+
+const baseTeamRoute = "api/v1/protected/teams/";
 
 const makeReservation = (blockId: number, teamId?: number): Promise<void> => {
   return AppAxiosInstance.post(ProtectedApiClientRoutes.MAKE_RESERVATION, {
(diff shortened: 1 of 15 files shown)


src/api/protectedApiClient.ts
```diff
@@ -165,6 +228,123 @@ const changePrivilegeLevel = (request: {
     .catch((e) => e);
 };
 
+const createTeam = (request: {
+  name: string;
+  bio: string;
+  inviteEmails: string[];
```
rymaju, line 234: ⛏️ This is the same type as on line 42, maybe we might want to declare an interface for this?

src/containers/availableTeams/ducks/selectors.ts
```diff
@@ -0,0 +1,18 @@
+import { AsyncRequest, AsyncRequestKinds } from '../../../utils/asyncRequest';
+import { AvailableTeam } from './types';
+import { TabItem } from '../../../components/leaderboard/types';
+
+export const availableTeamsToTabItems = (
+  availableTeams: AsyncRequest<AvailableTeam[], any>,
+): TabItem[] => {
+  if (availableTeams.kind === AsyncRequestKinds.Completed) {
```
rymaju, line 8: ⛏️ We might want to use AsyncRequestIsCompleted here instead of checking kind directly

src/containers/teamPage/ducks/selectors.ts
```diff
@@ -0,0 +1,45 @@
...
+      startDate: new Date(goal.startDate),
+      completeBy: new Date(goal.completeBy),
+      completionDate: !!goal.completeBy ? new Date(goal.completeBy) : null,
+    };
+  });
+};
+
+// In case the team request has not been completed.
+// This is to prevent the TeamProps from being undefined
+const emptyTeam: TeamProps = {
```
rymaju, line 39: I think this is going to refer to one object, if we mutate it that means this will also be mutated. Consider making a function that returns a fresh object every time instead or freezing it?
```suggestion
const emptyTeam: () => TeamProps = () => ({
  id: 0,
  name: '',
  bio: '',
  members: [],
  goals: [],
});
```

rymaju (commented):
Couple tiny things but otherwise looks pretty good to me 🥳

src/api/protectedApiClient.ts
```diff
@@ -33,6 +38,39 @@ export interface ProtectedApiClient {
     newLevel: string;
     password: string;
   }) => Promise<void>;
+  readonly createTeam: (request: {
+    name: string;
+    bio: string;
+    inviteEmails: string[];
+  }) => Promise<void>;
```
jackblanc, line 45: Is there an interface defined for this request object? If so we should reference it

jackblanc: same thing for other cases

src/api/protectedApiClient.ts
jackblanc: yes, something like `CreateTeamRequest`?

src/api/protectedApiClient.ts
```diff
@@ -165,6 +228,123 @@ const changePrivilegeLevel = (request: {
     .catch((e) => e);
 };
 
+const createTeam = (request: {
+  name: string;
+  bio: string;
+  inviteEmails: string[];
+}): Promise<void> => {
+  return AppAxiosInstance.post(ProtectedApiClientRoutes.CREATE_TEAM, request)
+    .then((r) => r.data)
```
jackblanc, line 237: Personally, I think it's better to use `res` and `err`, as it's more readable/grep-able.

src/containers/availableTeams/ducks/selectors.ts
jackblanc: agree, utilities should be preferred

src/containers/teamPage/ducks/selectors.ts
```diff
@@ -0,0 +1,45 @@
...
+export const teamResponseRequestToTeamProps = (
+  team: AsyncRequest<TeamResponse, any>,
+): TeamProps => {
+  if (team.kind === AsyncRequestKinds.Completed) {
+    return teamResponseToTeamProps(team.result);
+  }
+  return emptyTeam;
+};
+
+export const teamResponseToTeamProps = (team: TeamResponse): TeamProps => {
```
jackblanc, line 13: should both of these functions be publicly exposed?

src/containers/teamPage/ducks/selectors.ts
jackblanc: Yeah, good idea. Although, it may make more sense to alter wherever this is used to simply use these defaults locally when an undefined object is received.

src/containers/teamPage/ducks/types.ts
```diff
@@ -7,18 +13,18 @@ export interface TeamProps {
...
 export interface GoalProps {
   id: number;
   goal: number;
   progress: number;
-  start_date: Date;
-  complete_by: Date;
-  completion_date: Date | null;
+  startDate: Date;
+  completeBy: Date;
+  completionDate: Date | null;
```
jackblanc, line 27:
```suggestion
  completionDate?: Date;
```

src/containers/teamPage/ducks/types.ts
```diff
@@ -27,3 +33,48 @@ export enum TeamRole {
...
+
+export interface TeamResponse {
+  id: number;
+  name: string;
+  bio: string;
+  members: MemberProps[];
+  goals: GoalResponseJSON[];
+  finished: boolean;
+  createdAt: string;
+  deletedAt: string | null;
```
jackblanc, line 57:
```suggestion
  deletedAt?: string;
```

src/containers/teamPage/ducks/types.ts
```diff
@@ -27,3 +33,48 @@ export enum TeamRole {
...
+}
+
+export interface GoalResponseJSON {
+  goalId: number;
+  goal: number;
+  progress: number;
+  startDate: number;
+  completeBy: number;
+  completionDate: number | null;
```
jackblanc, line 66:
```suggestion
  completionDate?: number;
```

src/containers/teamPage/index.tsx
```diff
@@ -98,44 +98,44 @@ const Line = styled.div`
 // dummy data
```
jackblanc, line 98: good, I hate snake_case

jackblanc (approved):
Looks great! Left some suggestions

PR #51, opened 2021-03-30
Bug report footer

liammoyn:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1152597582)

Added a footer to the bottom of the site with a link to a bug reporting form.

## This PR

Added a custom footer in the App. Used the default AntD Footer component but styled to reduce the vertical padding and to center the content.

The footer is at the bottom of all content, in most pages you have to scroll to see it, including the map page (see the attached screenshots and notice the page is scrolled).

## Screenshots

Screen Shot 2021-03-30 at 5 17 14 PM
Screen Shot 2021-03-30 at 5 17 29 PM
Screen Shot 2021-03-30 at 5 17 31 PM

 

## Verification Steps

Checked that the link works and the footer looks reasonable on chrome, firefox, and safari

### DIFF
--- a/src/components/siteFooter/index.tsx
+++ b/src/components/siteFooter/index.tsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import { Layout, Typography } from 'antd';
+import { BugOutlined } from '@ant-design/icons';
+
+const { Footer } = Layout;
+const { Text, Link } = Typography;
+
+const SiteFooter: React.FC = () => {
+  return (
+    <>
+      <Footer>
+        <BugOutlined />
+        <Text> Notice something off? </Text>
+        <Link href="https://forms.gle/4dLLYX4z5MMsgaAq7" target="_blank">
+          Submit a Bug Report!
+        </Link>
+      </Footer>
+    </>
+  );
+};
+
+export default SiteFooter;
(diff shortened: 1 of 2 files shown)


src/components/siteFooter/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  margin: auto;
+`;
+
+const SiteFooter: React.FC = () => {
+  return (
+    <>
+      <StyledFooter>
+        <BugOutlined />
+        <Text> Notice something off? </Text>
+        <Link href="https://forms.gle/4dLLYX4z5MMsgaAq7" target="_blank">
```
SofieDunt, line 21: ⛏️ Might be nice to store the link under assets, but I don't think we'll use it anywhere else so this is probably good.

SofieDunt (approved):
Looks great! 🔥

jackblanc (changes requested):
Personally, I think in this case we should take more care to incorporate this into the page in a UI-friendly way. It makes sense to get something up ASAP on the prod LLB site, but here we should take time because nobody is using this yet. Let's check in with Brandon and see what he thinks, as this current approach changes the appearance of every page.

PR #52, opened 2021-03-31
Connect User Data

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/1114626135)
[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/1114627064)

Connects the userData route and uses it to personalize the site for users.

## This PR

- Added route to protectedApiClient
- Created selectors to get
  - first name, used in home page
  - full name, used in nav bar and settings page
  - email, used in settings page
- Dispatch to get user details after successful login (in login container, when privilege is not none)
- Changed settings form to take the new email and password to match change email route (other necessary changes for settings documented in a ticket)

## Screenshots

 
Not Jack in Home and NavBar
Home
Display name and email in Settings
Settings

## Verification Steps

 
- Added a test case for `getUserData`
- Ran frontend and backend locally, signed into different accounts, and ensured the data displayed correctly


jackblanc (commented):
This looks great, nice work on this! A few comments and suggestions:
- Could you add this to the frontend-scaffold when it's finished? Other projects will also need it. I can copy the code over as well if you'd like. 
- I _think_ we might want to consider adding this new field to the UserAuthenticationReducerState, as opposed to creating a new sub-state. The fields seem to be similar enough. 
- Going forwards, we need to be testing all Redux code (I like just decided this today so no way you should have known). I will give a presentation on this on Sunday or can pair today to help. There a few good examples already in this codebase.

src/components/navBar/index.tsx
```diff
@@ -132,7 +143,9 @@ const NavBar: React.FC<NavBarProps> = ({ tokens }) => {
     return (
       <FlexDiv>
         {/* This needs to changed, not a constant */}
```
willmt80, line 145: We shouldn't need this comment anymore

src/api/protectedApiClient.ts
```diff
@@ -165,6 +168,12 @@ const changePrivilegeLevel = (request: {
     .catch((e) => e);
 };
 
+const getUserData = (): Promise<UserData> => {
+  return AppAxiosInstance.get(ProtectedApiClientRoutes.GET_USER_DATA)
+    .then((r) => r.data)
+    .catch((e) => e);
```
jackblanc, line 174:
```suggestion
    .then((res) => res.data)
    .catch((err) => err);
```
⛏️

jackblanc (approved):
🚢

src/api/protectedApiClient.ts
rymaju: +1

src/components/navBar/index.tsx
```diff
@@ -131,8 +140,9 @@ const NavBar: React.FC<NavBarProps> = ({ tokens }) => {
   const LoggedInExtra = () => {
     return (
       <FlexDiv>
-        {/* This needs to changed, not a constant */}
-        <Paragraph style={{ margin: 'auto 20px auto 0' }}>Jack Blanc</Paragraph>
+        <Paragraph style={{ margin: 'auto 20px auto 0' }}>
```
rymaju, line 143: ⛏️ Its a little odd that this is isnt a styled component but since its just margins I dont think it really matters

rymaju (approved):
LGTM 👍

PR #57, opened 2021-04-21
Create Tree Popup

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/926131646)

Creates popup to let viewers see more information about a tree/navigate to that tree's page on the map.

## This PR

Created the popup component to be linked to the map as a [custom popup](https://developers.google.com/maps/documentation/javascript/examples/overlay-popup#all) (rather than an info window to allow for increased customization)
- tried using AntD's popover but this doesn't work well with the map for placement and toggling visibility
- `popRef` is for integration with the map

## Screenshots

 
Tree Popover
Tree Popover Zoom

## Verification Steps

 
Visually confirmed it matched the designs, placed it onto the map blocks layer to make sure it would work well once the sites are added

### DIFF
--- a/src/components/treePopup/index.tsx
+++ b/src/components/treePopup/index.tsx
@@ -1,0 +1,111 @@
+import React from 'react';
+import styled from 'styled-components';
+import { Typography } from 'antd';
+import { LinkButton } from '../linkButton';
+import { Routes } from '../../App';
+import {
+  BLACK,
+  DARK_GREY,
+  LIGHT_GREEN,
+  LIGHT_GREY,
+  MID_GREEN,
+  WHITE,
+} from '../../utils/colors';
+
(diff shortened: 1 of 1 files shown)


willmt80 (approved):
Looks good to me! 💯. Do you expect we will run into any trouble placing this correctly on top of trees?

SofieDunt:
Placing this on trees should be fairly straightforward - I imagined we would make a custom popup class to link this to the map and set its position based on an event listener for when a user clicks on a tree in the sites data layer (which is what I did to get those screenshots, but with blocks). There will probably be a few bugs to work out with visibility and references, but those have to do with the map loader more than this component.

jackblanc (approved):
Nice

PR #58, opened 2021-04-21
Keep Login/Signup Form Values on Resize

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/1012460652)

Prevents form fields from being cleared on window resize.

## This PR

- Used `Form.useForm` to keep form instance consistent across forms rendered on different window types
- Passed `windowType` to forms as a prop rather than having both pages and forms use the `useWindowDimensions` hook (less event listeners)

## Screenshots

 
No styling changes.

## Verification Steps

 
Ran locally, input values into form fields and resized the browser past each breakpoint to check that
- layout updated according to designs
- values did not clear past breakpoints

### DIFF
--- a/src/components/signupForm/index.tsx
+++ b/src/components/signupForm/index.tsx
@@ -12,7 +12,8 @@ import {
   FullWidthSpace,
   Gap,
 } from '../themedComponents';
-import useWindowDimensions, { WindowTypes } from '../windowDimensions';
+import { WindowTypes } from '../windowDimensions';
+import { FormInstance } from 'antd/es/form';
 
 const { Paragraph } = Typography;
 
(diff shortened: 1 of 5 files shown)


src/components/signupForm/index.tsx
```diff
@@ -24,18 +25,23 @@ const Footer: typeof Paragraph = styled(Paragraph)<ParagraphProps>`
...
+}) => {
   return (
     <>
-      <Form name="basic" initialValues={{ remember: true }} onFinish={onFinish}>
+      <Form name="basic" form={formInstance} onFinish={onFinish}>
         <FullWidthSpace direction="vertical" size={1}>
           <FormRow>
             <FormHalfItem
+              name="firstName"
               rules={[
```
willmt80, line 45: 💭  Maybe create a form rules file with all of the rules we use throughout the app.

willmt80 (changes requested):
Looks good! left a suggestion about reusing the form item rule logic

src/components/signupForm/index.tsx
SofieDunt: Yeah, I thought about it but the only rules in common between the login/signup form is the one for email. The signup form has more in common with the settings form, so I was just going to pull out the common ones in the ticket to fix settings. You're right though, I'll create a `formRules` file as part of this and update settings later. Do you think we should put all of the rules in there or just the ones used in multiple files?

jackblanc (approved):
👍

rymaju (approved):
Looks pretty straightforward 💯

willmt80:
💯

willmt80 (approved):

PR #60, opened 2021-04-28
Connect Sites Geojson Route

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/1247539229)

Connects get sites in geojson route so that site information can be used, ex. in displaying trees on the map.

## This PR

- Added route to ApiClientRoutes and tested successful request
- Added siteGeoData to the MapGeoDataReducerState, created necessary ducks, tested reducers and thunks
- Added sites to MapViewProps (setup for ticket to make the tree map)

## Verification Steps

 
Made sure unit tests passed

### DIFF
--- a/src/components/mapPageComponents/ducks/test/reducers.test.ts
+++ b/src/components/mapPageComponents/ducks/test/reducers.test.ts
@@ -1,0 +1,128 @@
+import { blockGeoData, neighborhoodGeoData, siteGeoData } from '../actions';
+import reducers, { initialMapGeoDataState } from '../reducer';
+import {
+  BlockGeoData,
+  NeighborhoodGeoData,
+  SiteGeoData,
+  MapGeoDataReducerState,
+} from '../types';
+import { AsyncRequestCompleted } from '../../../../utils/asyncRequest';
+
+describe('Map Reducers', () => {
+  describe('BlockGeoData', () => {
+    it('Updates state correctly when block data is retrieved', () => {
+      const blockData: BlockGeoData = {
+        type: 'FeatureCollection',
+        name: 'blocks',
+        features: [
+          {
+            type: 'Feature',
+            properties: {
+              blockId: 1714,
+              lat: 42.3488784985,
+              lng: -71.0293810011,
+            },
+            geometry: {
+              type: 'MultiPolygon',
+              coordinates: [
+                [
+                  [
+                    [-71.02859399925286, 42.34889700150278],
+                    [-71.02912299801895, 42.34837799993562],
+                  ],
+                ],
+              ],
+            },
+          },
+        ],
+      };
+      const action = blockGeoData.loaded(blockData);
+      const expectedNextState: MapGeoDataReducerState = {
+        ...initialMapGeoDataState,
+        blockGeoData: AsyncRequestCompleted<BlockGeoData, void>(blockData),
+      };
+      expect(reducers(initialMapGeoDataState, action)).toEqual(
+        expectedNextState,
+      );
+    });
+  });
+
+  describe('NeighborhoodGeoData', () => {
+    it('Updates state correctly when neighborhood data is retrieved', () => {
+      const neighborhoodData: NeighborhoodGeoData = {
+        type: 'FeatureCollection',
+        name: 'neighborhoods',
+        features: [
+          {
+            type: 'Feature',
+            properties: {
+              neighborhoodId: 15,
+              name: 'Roslindale',
+              completionPerc: 0,
+              lat: 42.2803,
+              lng: -71.1266,
+            },
+            geometry: {
+              type: 'MultiPolygon',
+              coordinates: [
+                [
+                  [
+                    [-71.12592717485386, 42.272013107957406],
+                    [-71.12610933458738, 42.2716219294518],
+                  ],
+                ],
+              ],
+            },
+          },
+        ],
+      };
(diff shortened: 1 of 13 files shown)


src/components/mapPageComponents/ducks/types.ts
```diff
@@ -46,18 +46,45 @@ interface NeighborhoodFeaturePropertiesResponse {
...
+  lat: number;
+  lng: number;
+}
+
 // ---------------------------------Shared Types----------------------------------------
 
 interface MapGeometry {
   type: string;
-  coordinates: Coordinate[][][];
+  coordinates: Coordinate[][][] | Coordinate;
```
rymaju, line 79: ⛏️ This data type seems a little strange? Do you think you might want some tsdoc to explain why it can be a  List of List of List of Coordinate or just a coordinate? It seems a bit strange

src/containers/landing/index.tsx
```diff
@@ -18,20 +18,22 @@ import { C4CState } from '../../store';
...
+    sites: state.mapGeoDataState.siteGeoData,
   };
 };
 
 const PaddedContent = styled.div`
   padding: 24px 50px;
 `;
 
-const Landing: React.FC<LandingProps> = ({ neighborhoods, blocks }) => {
+const Landing: React.FC<LandingProps> = ({ neighborhoods, blocks, sites }) => {
```
rymaju, line 36: ❓ For all of these request props, this all looks good, but I only see code for the `asyncRequestIsComplete` state. What happens when any of these requests fail or are loading? Are we displaying that information to the user with a error message alert or spinner?

rymaju (approved):
Left some general questions, looks like this works, but I wonder about the error and loading states for some of this data, and I'm also a bit confused by the types.

src/components/mapPageComponents/ducks/test/reducers.test.ts
```diff
@@ -0,0 +1,128 @@
+import { blockGeoData, neighborhoodGeoData, siteGeoData } from '../actions';
```
willmt80, line 1: 🔥

src/components/mapPageComponents/ducks/test/thunks.test.ts
```diff
@@ -0,0 +1,180 @@
+import { BlockGeoData, NeighborhoodGeoData, SiteGeoData } from '../types';
```
willmt80, line 1: 🔥 🔥 🔥

src/components/mapPageComponents/mobileMapPage/index.tsx
```diff
@@ -14,18 +14,24 @@ const MainContent = styled.div`
...
   neighborhoods,
+  sites,
   children,
 }) => (
   <>
     <MainContent>
       <PageLayout>
         <Content>
-          <MapView blocks={blocks} neighborhoods={neighborhoods} />
+          <MapView
```
willmt80, line 30: ❓ Will the blocks and the sites both be passed to the map view on the reservation and tree maps?

src/components/mapPageComponents/ducks/test/thunks.test.ts
```diff
@@ -0,0 +1,180 @@
...
+          getNeighborhoodGeoData: mockGetNeighborhoodGeoData,
+          getSiteGeoData: mockGetSiteGeoData,
+        },
+      };
+
+      await getMapGeoData()(mockDispatch, getState, mockExtraArgs);
+
+      expect(mockDispatch).toHaveBeenCalledTimes(6);
+      expect(mockDispatch).toHaveBeenNthCalledWith(
+        4,
```
willmt80, line 119: ❓ Not super familiar with these kinds of tests. What does this number signify?

willmt80 (approved):
Looks Great! Left a couple questions in addition to Ryan's comments.

src/components/mapPageComponents/ducks/types.ts
SofieDunt: This is because GeoJSON has different [coordinates types](https://tools.ietf.org/html/rfc7946#section-3.1.1) depending on the geometry type, added a reference comment.

src/containers/landing/index.tsx
SofieDunt: No, as is the map view just wouldn't load. I'll open a separate ticket to handle these cases.

src/components/mapPageComponents/mobileMapPage/index.tsx
SofieDunt: Yeah, I was thinking the map view would have all the data and we'd toggle visibility based on another prop indicating map view. This is a good point though, for adding the tree map I'll see if it's easier to just not pass one or the other.

src/components/mapPageComponents/ducks/test/thunks.test.ts
SofieDunt: These probably explain it better than I could:
- [toHaveBeenCalledTimes](https://jestjs.io/docs/expect#tohavebeencalledtimesnumber)
- [toHaveBeenNthCalledWith](https://jestjs.io/docs/expect#tohavebeennthcalledwithnthcall-arg1-arg2-)

but basically I'm checking that the mock get map geo data dispatch was called 6 times, and on the 4th call it was called with `blockGeoData.loaded(mockBlockDataResponse)` as the argument (the first three calls are the different `.loading` states).

PR #61, opened 2021-04-30
Add modal

vrushalitarte01:
## Why

 Just updates index.tsx in mapView so it includes releaseReservation for the case where the block clicked is reserved (in addition to makeReservation being called for the case where the block clicked is open).


src/components/mapPageComponents/mapView/index.tsx
```diff
@@ -43,14 +43,15 @@ const MapView: React.FC = () => {
...
+        protectedApiClient.makeReservation(activeBlockId); 
         break;
       case ReservationModalType.RESERVED:
-        //set block status to open
+        //release block
+        protectedApiClient.releaseReservation(activeBlockId);
         break;
       default:
-        //block clicked not owned/open
+        //block clicked not owned/open so do nothing
```
willmt80, line 54: Will this condition ever be hit? if so we should try to disable any buttons or interactions that would allow this.

willmt80 (commented):
Looks good for the most part! Just left a question about the default case

vrushalitarte01:
changes covered in addTeams

PR #64, opened 2021-05-01
Update Settings

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/pulses/1175092158)

Updates the settings form so that users can change their username, email, and password and delete their account (TODO: need to look into this more, backend was giving a successful user deleted response but didn't update `deleted_at`) and sees messages confirming updates or informing them of errors. Previously form fields didn't match requests, and errors couldn't be seen. Also creates a mobile layout for improved user experience.

## This PR

ProtectedApiClient:
- updated `DELETE_USER` route to match API Specification
- updated changePassword, changeUsername, changeEmail, and deleteUser so that errors can be accessed in the forms to display error message

Forms:
- created separate form components for each form, improves readability and reusability for different window types
- created interfaces for each request
- change email form gets user data again after update to update info displayed on settings page

Settings:
- delete user as a modal (extra confirmation that user wants to delete account)
- mobile view as a scroll

Signup:
- get user data after signup

## Screenshots

 
Desktop (same layout for narrow desktop and tablet)
DesktopSettings
Delete Account Modal
DeleteAccountModal
Messages
ErrorMessageDesktop
SuccessMessageDesktop
Mobile
Mobile1
Mobile2

## Verification Steps

 
Tested all forms and checked that updates were successful/correct error messages received. Made sure layouts looked reasonable across window types.

### DIFF
--- a/src/containers/settings/index.tsx
+++ b/src/containers/settings/index.tsx
@@ -69,7 +69,7 @@ const Settings: React.FC<SettingsProps> = ({ tokens, userData }) => {
       <Helmet>
         <title>Settings</title>
         <meta
-          name="Settings"
+          name="description"
           content="Where the user can change their account settings, including their username, email, and password, and delete their account."
         />
       </Helmet>
--- a/src/containers/settings/index.tsx
+++ b/src/containers/settings/index.tsx
@@ -1,31 +1,41 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { connect, useSelector } from 'react-redux';
-import { Col, Row, Button, Form, Input, Typography } from 'antd';
+import { Col, Row, Button, Typography, Modal } from 'antd';
 import PageHeader from '../../components/pageHeader';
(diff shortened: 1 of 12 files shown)


src/containers/settings/index.tsx
```diff
@@ -45,163 +55,115 @@ const UserInformationText = styled(Paragraph)`
...
-                    },
-                    {
-                      min: 8,
-                      message: passwordRules,
-                    },
-                  ]}
+      <Helmet>
+        <title>Settings</title>
+        <meta
+          name="Settings"
```
rymaju, line 72: ⛏️ This meta tag doesnt do anything, there is no meta tag called "Settings", I think this is supposed to be a "description" meta tag

rymaju, line 72:
```suggestion
          name="description"
```

rymaju (approved):
One small thing with the meta tag, otherwise looks pretty good!

PR #65, opened 2021-05-02
Force Refresh on Logout

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1258029135)

Forces refresh on logout so users are auth-redirected to login and can't see user data of the last person logged in.

## This PR

Force refresh after dispatching logout with `history.go(0)`. 

## Verification Steps

 
Logged in and tested logout on various pages.


rymaju (approved):
LGTM, but if you still want to go to the landing page you should push to that route in addition to refreshing.

PR #66, opened 2021-05-02
Map Data Request, WindowType, and isLoggedIn Handling

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1258024804)
[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1175108080)

- Informs user when map data is loading/failed to load (while loading/failed, still able to view non-map components on the page)
- Only shows login/signup buttons on mobile landing bar if the user isn't signed in
- Improves reusability of components

## This PR

MapContent
- Created new component to access map geo data (rather than having to access and check request type across all containers that want to use the map)
- Checks if the map data requests are loaded, loading, or failed and displays map, spin, or error accordingly

MobileLandingBar
- Pass whether the user is logged in as a prop

PageHeader/MobilePageHeader
- Deleted MobilePageHeader, PageHeader takes `isMobile` as a prop instead (I find this cleaner)

isCheck
- Created utils to check types (isLoggedIn and isMobile)

General
- Used utils where possible

## Screenshots

 
Loading
LoadingDesktop
LoadingMobile
Error
ErrorDesktop
No Login/Signup Buttons
NoButtons

## Verification Steps

 
Ran locally and checked map display and styling. Tested isCheck functions.


src/components/mapPageComponents/mobileLandingBar/index.tsx
```diff
@@ -64,17 +65,19 @@ const SignUpButton = styled(Button)`
 
 const LandingStatsContainer = styled.div`
   width: 90vw;
-  margin: 0px -30px;
+  margin: -3vh -30px;
```
rymaju, line 68: ❓ Negative margin is questionable in general, but if it works it works 🤷

src/components/pageHeader/index.tsx
```diff
@@ -28,27 +28,37 @@ const StyledSubtitle = styled(Paragraph)`
...
-        }}
-      >
-        {pageTitle}
-      </StyledTitle>
-      <StyledSubtitle textColor={subtitleColor}>{pageSubtitle}</StyledSubtitle>
-    </div>
-  );
+  if (isMobile) {
+    return (
+      <div>
```
rymaju, line 48: ⛏️ These `<div>` elements can be react fragments, which might make the final DOM tree look a little nicer.

src/containers/landing/index.tsx
```diff
@@ -1,40 +1,34 @@
...
+import styled from 'styled-components';
+import MobileMapPage from '../../components/mapPageComponents/mobileMapPage';
+import MobileLandingBar from '../../components/mapPageComponents/mobileLandingBar';
 
 const PaddedContent = styled.div`
   padding: 24px 50px;
 `;
 
-const Landing: React.FC<LandingProps> = ({ neighborhoods, blocks, sites }) => {
+interface LandingProps {
```
rymaju, line 23: 🏕️ You dont actually use LandingProps anymore so you can delete this and the connect state to props at the bottom of this component

src/utils/isCheck.ts
```diff
@@ -0,0 +1,18 @@
+import { PrivilegeLevel } from '../auth/ducks/types';
+import { WindowTypes } from '../components/windowDimensions';
+
+/**
+ * Checks if the user is logged in
+ * @param privilegeLevel the user's privilege level
+ */
+export function isLoggedIn(privilegeLevel: PrivilegeLevel): boolean {
```
rymaju, line 8: 🤔 I wonder if this might be better as a selector on auth? Maybe not but just a thought

rymaju (approved):
Theres some redundant props, but overall looks pretty solid.

src/utils/isCheck.ts
SofieDunt: Good call

PR #68, opened 2021-05-06
Connect Tree Popup

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1224885844)

Adds tree popup to map so users can easily navigate to tree pages.

## This PR

MapView
- created a [custom popup](https://developers.google.com/maps/documentation/javascript/examples/overlay-popup)
- changed `useEffect` dependencies so the map doesn't reload whenever the popup is added/popup data is changed (or whenever the reservation modal is shown) by
  - making `markersArray` a constant within `useEffect`
  - making the dependencies `mapRef.current` and `treePopupRef.current` rather than the refs themselves. Since the refs update after DOM loads, created useEffects to update `mapElement` and `treePopupElement` as needed
 
TreePopup
- is not visible by default (otherwise would show on map before click event) and may be closed
- capitalizes species names, adjusts appearance if tree does not have a species name or an address
- overflow for long addresses, max-height so popups with little content are smaller

## Screenshots

 
AllData
NoAddress
NoSpecies

## Verification Steps

 
Interacted with map, made sure:
- tree popups appeared on/when site clicked
- did not appear on other clicks/at any other time
- popups linked to appropriate tree page

### DIFF
--- a/src/components/mapPageComponents/mapView/index.tsx
+++ b/src/components/mapPageComponents/mapView/index.tsx
@@ -90,7 +90,7 @@ const MapView: React.FC<MapViewProps> = ({
   const { windowType } = useWindowDimensions();
 
   const mapRef = createRef<HTMLDivElement>();
-  const treePopoverRef = createRef<HTMLDivElement>();
+  const treePopupRef = createRef<HTMLDivElement>();
 
   const loader = new Loader({
     apiKey: process.env.REACT_APP_GOOGLE_MAPS_KEY || '',
@@ -102,8 +102,8 @@ const MapView: React.FC<MapViewProps> = ({
   const mapId = '76c08a2450c223d9';
 
   const [mapElement, setMapElement] = useState(mapRef.current);
-  const [treePopoverElement, setTreePopoverElement] = useState(
-    treePopoverRef.current,
+  const [treePopupElement, setTreePopupElement] = useState(
+    treePopupRef.current,
   );
 
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


src/components/mapPageComponents/mapView/index.tsx
```diff
@@ -109,6 +131,49 @@ const MapView: React.FC<MapViewProps> = ({
           },
         });
 
+        class Popup extends google.maps.OverlayView {
```
rymaju, line 134: ❓ I'm not sure about putting a class definition in the body of a function. This really deserves its own file I think.

src/components/mapPageComponents/mapView/index.tsx
```diff
@@ -61,6 +63,13 @@ const MapView: React.FC<MapViewProps> = ({
   );
```
rymaju, line 63: ❗ Wow there are over 400 lines of code in this react component excluding JSX. That is _way_ too much. Almost all of this should be extracted into another file.

rymaju (approved):
Looks solid, I think the mapView page definitely needs a lot of refactoring though. Currently ALL of the code in the function body will be rerun on any rerender. This looks like a pretty bad performance problem. We should meet to look into solutions sometime.

src/components/mapPageComponents/mapView/index.tsx
SofieDunt: Yeah, I'll look more into this. The problem with pulling the class out (and most of the code in this hook) is that `google` isn't defined until the DOM loads.

src/components/mapPageComponents/mapView/index.tsx
SofieDunt: Agreed, I'll open a ticket for cleanup

PR #70, opened 2021-05-06
connecting stewardship and get site routes

willmt80:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1185039378)

This PR allows users to view the stats about a particular tree as well as the stewardship activities that have been performed. This also allows the user to adopt, unadopt, and perform stewardship activities on those trees

## This PR

Created the api routes, reducers, thunks, and selectors that retrieve data from the backend, put it in redux, and display that data on the frontend. I also made changes to the treepage that allowed it to work with these changes

## Screenshots

 

logged out tree

real tree

## Verification Steps

 

Double checked the results of these routes in postman. Made sure those calls gave the same results from the UI. Currently the unadopt does not work in the backend so once that is fixed there may be some fixes needed. The get routes have results that are displayed on the page so those are fairly self evident. Wrote tests for the routes. Will make another ticket to complete the thunks, selector, and reducer tests. This PR is just somewhat time sensitive.

### DIFF
--- a/src/containers/treePage/index.tsx
+++ b/src/containers/treePage/index.tsx
@@ -1,14 +1,12 @@
 import React, { useEffect } from 'react';
 import { useHistory } from 'react-router-dom';
-import { useParams } from 'react-router-dom';
+import { useParams, useLocation } from 'react-router-dom';
 import PageLayout from '../../components/pageLayout';
 import ReturnButton from '../../components/returnButton';
 import { Row, Col, Typography, Button, Card, message, Form } from 'antd';
-import { Routes } from '../../App';
+import { RedirectStateProps, Routes } from '../../App';
 import { Helmet } from 'react-helmet';
-import {
-  UserAuthenticationReducerState,
-} from '../../auth/ducks/types';
+import { UserAuthenticationReducerState } from '../../auth/ducks/types';
 import { isLoggedIn } from '../../auth/ducks/selectors';
 import { TreeCare, ActivityRequest } from './ducks/types';
 import PageHeader from '../../components/pageHeader';
@@ -114,6 +112,7 @@ const TreePage: React.FC<TreeProps> = ({ siteData, stewardShip, tokens }) => {
   const dispatch = useDispatch();
   const id = Number(useParams<TreeParams>().id);
   const history = useHistory();
+  const location = useLocation<RedirectStateProps>();
 
   const [stewardshipFormInstance] = Form.useForm();
 
@@ -223,22 +222,23 @@ const TreePage: React.FC<TreeProps> = ({ siteData, stewardShip, tokens }) => {
                             Unadopt
                           </Button>
                         ) : (
-                          <Button
-                            type="primary"
-                            size="large"
-                            onClick={onClickAdopt}
-                          >
-                            Adopt
-                          </Button>
+                          <>
+                            <Button
+                              type="primary"
+                              size="large"
+                              onClick={onClickAdopt}
+                            >
+                              Adopt
+                            </Button>
+                            <StewardshipContainer>
+                              <Title level={3}>Record Tree Care</Title>
+                              <StewardshipForm
+                                onFinish={onFinishRecordStewardship}
+                                form={stewardshipFormInstance}
+                              />
+                            </StewardshipContainer>
+                          </>
                         )}
-
-                        <StewardshipContainer>
-                          <Title level={3}>Record Tree Care</Title>
-                          <StewardshipForm
-                            onFinish={onFinishRecordStewardship}
-                            form={stewardshipFormInstance}
-                          />
-                        </StewardshipContainer>
                       </>
                     ) : (
                       <>
@@ -246,7 +246,11 @@ const TreePage: React.FC<TreeProps> = ({ siteData, stewardShip, tokens }) => {
                         <Button
                           type="primary"
                           size={'large'}
-                          onClick={() => history.push(Routes.LOGIN)}
+                          onClick={() =>
+                            history.push(Routes.LOGIN, {
+                              destination: location.pathname,
+                            })
+                          }
                         >
                           Log In
                         </Button>
--- a/src/containers/treePage/index.tsx
+++ b/src/containers/treePage/index.tsx
@@ -10,7 +10,7 @@ import {
   UserAuthenticationReducerState,
 } from '../../auth/ducks/types';
 import { getPrivilegeLevel } from '../../auth/ducks/selectors';
-import { SiteProps, SiteEntry, TreeCare, SiteEntryNames, Activity, ActivityRequest } from './ducks/types';
+import { TreeCare, ActivityRequest } from './ducks/types';
 import PageHeader from '../../components/pageHeader';
 import { TitleProps } from 'antd/lib/typography/Title';
 import StewardshipForm from '../../components/stewardshipForm';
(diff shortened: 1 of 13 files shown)


src/containers/treePage/index.tsx
```diff
@@ -79,65 +99,78 @@ const EntryMessage = styled(Paragraph)`
...
+  const [stewardshipFormInstance] = Form.useForm();
+
+  useEffect(() => {
+    dispatch(getSiteData(id));
+    if (asyncRequestIsComplete(tokens)) {
+      dispatch(getAdoptedSites());
+    }
+  }, [dispatch, id, tokens]);
+
+  const onFinishRecordStewardship = (values: {
```
SofieDunt, line 126: ❓ Is it better for this to be here rather than in the form component?

src/containers/treePage/index.tsx
```diff
@@ -79,65 +99,78 @@ const EntryMessage = styled(Paragraph)`
...
+      weeded: values.stewardshipActivities.includes('Weeded'),
+    };
+    protectedApiClient
+      .recordStewardship(id, activities)
+      .then(() => {
+        message.success('Stwardship Recorded');
+        stewardshipFormInstance.resetFields();
+        dispatch(getSiteData(id));
+      })
+      .catch((err) =>
```
SofieDunt, line 144: ❓ Does this work? I thought the catch in protectedApiClient prevented us from actually catching the error here.

src/containers/treePage/index.tsx
```diff
@@ -79,65 +99,78 @@ const EntryMessage = styled(Paragraph)`
...
-  const latestEntry: SiteEntry = entries[entries.length - 1];
+  const onClickAdopt = () => {
+    protectedApiClient.adoptSite(id);
+    dispatch(getAdoptedSites());
+  };
+
+  const onClickUnadopt = () => {
+    protectedApiClient.unadoptSite(id);
+    dispatch(getAdoptedSites());
+  };
```
SofieDunt, line 157: Should we add error feedback to `onClickAdopt` and `onClickUnadopt`?

src/containers/treePage/index.tsx
```diff
@@ -79,65 +99,78 @@ const EntryMessage = styled(Paragraph)`
...
+      .catch((err) =>
+        message.error(`Failed to record stewardship: ${err.response.data}`),
+      );
+  };
 
-const TreePage: React.FC<SiteProps> = ({ address, entries }) => {
-  const latestEntry: SiteEntry = entries[entries.length - 1];
+  const onClickAdopt = () => {
+    protectedApiClient.adoptSite(id);
+    dispatch(getAdoptedSites());
```
SofieDunt, line 151: 💯

src/containers/treePage/index.tsx
```diff
@@ -79,65 +99,78 @@ const EntryMessage = styled(Paragraph)`
...
+    protectedApiClient.unadoptSite(id);
+    dispatch(getAdoptedSites());
+  };
+
+  const privilegeLevel: PrivilegeLevel = useSelector((state: C4CState) => {
+    return getPrivilegeLevel(state.authenticationState.tokens);
+  });
+
+  const doesUserOwnTree: boolean = useSelector((state: C4CState) => {
+    if (privilegeLevel !== PrivilegeLevel.NONE) {
```
SofieDunt, line 164: Can use `isLoggedIn` selector here

src/containers/treePage/index.tsx
```diff
@@ -149,62 +182,99 @@ const TreePage: React.FC<SiteProps> = ({ address, entries }) => {
         />
       </Helmet>
       <PageLayout>
-        <TreePageContainer>
-          {/*Change to tree map route once that page is finished*/}
-          <ReturnButton to={Routes.HOME}>{`<`} Return to Tree Map</ReturnButton>
-          <TreeMainContainer>
+        {asyncRequestIsComplete(siteData) && (
+          <TreePageContainer>
+            {/*Change to tree map route once that page is finished*/}
```
SofieDunt, line 187: Right now this is just the landing, do you think we should have a separate page for a tree map too?

src/containers/treePage/index.tsx
```diff
@@ -149,62 +182,99 @@ const TreePage: React.FC<SiteProps> = ({ address, entries }) => {
...
+              <Row>
+                <Col span={17}>
+                  <TreeInfoContainer>
+                    {siteData.result.entries[0].commonName && (
+                      <PageHeader
+                        pageTitle={siteData.result.entries[0].commonName}
+                      />
+                    )}
+                    <Title level={2}>{siteData.result.address}</Title>
+                    {privilegeLevel !== PrivilegeLevel.NONE ? (
```
SofieDunt, line 201: `isLoggedIn` selector

src/containers/treePage/index.tsx
```diff
@@ -149,62 +182,99 @@ const TreePage: React.FC<SiteProps> = ({ address, entries }) => {
...
+                        <StewardshipContainer>
+                          <Title level={3}>Record Tree Care</Title>
+                          <StewardshipForm
+                            onFinish={onFinishRecordStewardship}
+                            form={stewardshipFormInstance}
+                          />
+                        </StewardshipContainer>
+                      </>
+                    ) : (
+                      <Paragraph>Log in to adopt this tree!</Paragraph>
```
SofieDunt, line 230: ⛏️ Maybe a link to login? Would also allow redirect back to this site after logging in

src/containers/treePage/index.tsx
```diff
@@ -149,62 +182,99 @@ const TreePage: React.FC<SiteProps> = ({ address, entries }) => {
...
 
-const mapStateToProps = (): SiteProps => {
-  return dummyTree;
+const mapStateToProps = (state: C4CState): TreeProps => {
+  return {
+    tokens: state.authenticationState.tokens,
+    siteData: state.siteState.siteData,
+    stewardShip: mapStewardshipToTreeCare(
+      state.siteState.stewarshipActivityData,
+    ),
```
SofieDunt, line 275: 👍

src/store.ts
```diff
@@ -89,6 +109,8 @@ export const initialStoreState: C4CState = {
   mapGeoDataState: initialMapGeoDataState,
   teamState: initialTeamState,
   availableTeamsState: initialAvailableTeamsState,
+  siteState: initialSiteState,
+  adoptedSitesState: initialProtectedSiteState,
```
SofieDunt, line 113: ❓ Would it make sense to have `adoptedSites` be a field of `SiteReducerState` instead of having a separate sub-state?

SofieDunt (commented):
Looks great, thanks for doing this! Left some questions

src/containers/treePage/index.tsx
willmt80: The success case has a dispatch to get site data. This maps to data on other components on the tree page so at least for me it makes sense to keep it here.

src/containers/treePage/index.tsx
willmt80: True, I can remove this, however I'm not sure what makes the most sense to handle possible errors

src/containers/treePage/index.tsx
willmt80: Same thing as above, should we no longer catch the error in the api client?

src/containers/treePage/index.tsx
willmt80: No, moving this to the landing is fine

src/store.ts
willmt80: The adopted sites is a protected route while the other stewardship activities and site data are not. Meaning we want to be able to dispatch the public site actions without dispatching the adopted sites if the user is not logged in. Also because they're in different api clients have to have separate thunk actions (at least as far as I know).

src/containers/treePage/index.tsx
willmt80: Not super familiar with router and redirecting, right now the user is redirected to the home page. Do you know how to redirect the user back to this page after login?

src/containers/treePage/index.tsx
```diff
@@ -149,62 +197,107 @@ const TreePage: React.FC<SiteProps> = ({ address, entries }) => {
...
+                          </Button>
+                        )}
+
+                        <StewardshipContainer>
+                          <Title level={3}>Record Tree Care</Title>
+                          <StewardshipForm
+                            onFinish={onFinishRecordStewardship}
+                            form={stewardshipFormInstance}
+                          />
+                        </StewardshipContainer>
```
SofieDunt, line 241: This form should only be shown if the user owns the tree

src/containers/treePage/index.tsx
SofieDunt: I think you would use something like `history.push(Routes.LOGIN, { destination: currentLocation })`. If you push a location with a state object with `RedirectStateProps`, login can get the desired destination.

SofieDunt (approved):
💯

PR #71, opened 2021-05-07
reservation modal supports choosing teams for reservations

vrushalitarte01:
## Why

Users can now choose what team they want to reserve open blocks for.

## This PR

The handleOk function for the reservation modal now takes an optional team parameter which is passed to the makeReservation function and the function for the content of the modal for if the block is open now supports the option to map through the response for getTeams which returns an object that contains a list of teams to get team names for the user to select.

-- Calling reserve function returns malformed json request body error so need to check on that but modal itself works.

## Screenshots

 

Screen Shot 2021-05-07 at 10 54 41 AM

## Verification Steps

 

I made a test team and user and made the test user part of the test team and ensured that when the test user is signed in and clicks on an open block, is able to select the name for the test team and the id for the team selected is then passed to the makeReservation function.


src/components/mapPageComponents/mapView/index.tsx
```diff
@@ -46,18 +46,21 @@ const MapView: React.FC<MapViewProps> = ({ blocks, neighborhoods, sites }) => {
   // block id for modal
   const [activeBlockId, setActiveBlockId] = useState<number>(-1);
   // logic for reservation modal to complete action selected by user
-  const handleOk = (): void => {
+  const handleOk = (team?: number) => {
+    //check if team defined, if not throw error
```
SofieDunt, line 50: ❓ Where is the error thrown?

src/components/mapPageComponents/mapView/index.tsx
```diff
@@ -46,18 +46,21 @@ const MapView: React.FC<MapViewProps> = ({ blocks, neighborhoods, sites }) => {
   // block id for modal
   const [activeBlockId, setActiveBlockId] = useState<number>(-1);
   // logic for reservation modal to complete action selected by user
-  const handleOk = (): void => {
+  const handleOk = (team?: number) => {
```
SofieDunt, line 49: Would be nice to move `handleOk` into the reservation modal file and have `showModal` be a state of the reservation modal that updates to true whenever a new block ID is passed using `useEffect` - there's a lot of code in `MapView` that we want to cut down and move elsewhere

src/components/mapPageComponents/mapView/index.tsx
```diff
@@ -46,18 +46,21 @@ const MapView: React.FC<MapViewProps> = ({ blocks, neighborhoods, sites }) => {
...
+        protectedApiClient.makeReservation(activeBlockId, team);
+        protectedApiClient.completeReservation(activeBlockId, team);
         break;
       case ReservationModalType.RESERVED:
         // set block status to open
+        protectedApiClient.releaseReservation(activeBlockId);
         break;
-      default:
-        // block clicked not owned/open
+      case ReservationModalType.TAKEN:
```
SofieDunt, line 62: 👍

src/components/mapPageComponents/mapView/index.tsx
```diff
@@ -146,11 +149,11 @@ const MapView: React.FC<MapViewProps> = ({ blocks, neighborhoods, sites }) => {
             let color = 'green';
 
             // Use this for coloring reserved/completed blocks a different color
-            if (feature.getProperty('ID') % 10 === 0) {
+            if (feature.getProperty('block_id') % 10 === 0) {
```
SofieDunt, line 152: 💭 I wonder if we should have a `BlockProperties` enum (for neighborhoods and sites too) with all of the property names. Would make it easier to keep these consistent

src/components/reservationModal/index.tsx
```diff
@@ -43,15 +55,58 @@ const ReservationModal: React.FC<ReservationModalProps> = ({
...
-            <b>{`Are you sure?`}</b>
-            <br /> {`You want to reserve block ${blockID}`}
-          </text>
+          <>
+            {asyncRequestIsLoading(teamsRequest) && <Spin />}
+            {asyncRequestIsComplete(teamsRequest) && (
+              <>
+                <b>{`Are you sure?`}</b>
+                <br /> {`You want to reserve block ${blockID}`}
+                <br></br>
```
SofieDunt, line 84: ⛏️ Can do `<br />`

SofieDunt (commented):
Looks great! I think we'll want to move `handleOk` and `showModal` state into the reservation modal component itself to clean up `MapView`

PR #72, opened 2021-05-07
Reservation block and team id is case sensitive

willmt80:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1276131280)

The reservation requests no longer throw 400 errors because the requests have the correct case 

## This PR

just changed all the instances of "_id" to "ID"

## Screenshots

 

No styling

## Verification Steps

 

Confirmed with Vrushali that this fixed the error


vrushalitarte01 (approved):
confimed works

PR #73, opened 2021-05-23
made subtitlecolor prop lower case to remove react warning

willmt80:
## Why

There was a console warning about using camel case in dom props. This change removes that warning

## This PR

I simply changed the subtitle prop from camelCase to lowercase

## Screenshots

 

console

That particular warning is gone

## Verification Steps

 

The pageheaders still display their subtitles with the correct colors


SofieDunt (approved):
Thanks for doing this!

PR #74, opened 2021-05-25
added tree colors to map

willmt80:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1263481139)

This PR gives different colors to trees based on their attributes. Young trees are blue and adopted trees are red

## This PR

I added the planting date and adopter id to the site feature type and added a condition to the styling of the sites on the map that selects the correct color

## Screenshots

 

tree-colors

## Verification Steps

 

I used postman to make sure the planting date and adopter id values were coming back as expected. The changes are cosmetic so I was able to see the change in colors on the map

### DIFF
--- a/src/components/mapPageComponents/ducks/types.ts
+++ b/src/components/mapPageComponents/ducks/types.ts
@@ -70,6 +70,8 @@ interface SiteFeaturePropertiesResponse {
   address: string;
   lat: number;
   lng: number;
+  plantingDate?: number;
+  adopterId?: string;
 }
 
 // ---------------------------------Shared Types----------------------------------------
--- a/src/components/mapPageComponents/ducks/types.ts
+++ b/src/components/mapPageComponents/ducks/types.ts
@@ -63,11 +63,11 @@ interface SiteFeatureResponse {
 interface SiteFeaturePropertiesResponse {
   id: number;
   treePresent: boolean;
-  diameter: number;
-  species: string;
+  diameter?: number;
+  species?: string;
   updatedAt: string;
   updatedBy: string;
-  address: string;
+  address?: string;
   lat: number;
   lng: number;
   plantingDate?: number;
--- a/src/components/mapPageComponents/mapView/index.tsx
+++ b/src/components/mapPageComponents/mapView/index.tsx
@@ -374,8 +374,8 @@ const MapView: React.FC<MapViewProps> = ({
               visible = v && true;
             }
 
-            // TODO: update this to if the tree was planted within the past three years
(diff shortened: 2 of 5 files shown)


src/components/mapPageComponents/ducks/types.ts
```diff
@@ -70,6 +70,8 @@ interface SiteFeaturePropertiesResponse {
   address: string;
   lat: number;
   lng: number;
+  plantingDate?: number;
+  adopterId?: string;
```
SofieDunt, line 74: ❓ Do these have to be optional? Some of these other properties might also be null (ex. `address`), so should those be optional too? Either way, seems best to keep it consistent

src/components/mapPageComponents/mapView/index.tsx
```diff
@@ -375,9 +376,14 @@ const MapView: React.FC<MapViewProps> = ({
 
             // TODO: update this to if the tree was planted within the past three years
             // If the tree has not been updated within the past three years, use youngTreeIcon
```
SofieDunt, line 378: 🏕️ Can remove this todo and update the comment now

src/components/mapPageComponents/mapView/index.tsx
```diff
@@ -375,9 +376,14 @@ const MapView: React.FC<MapViewProps> = ({
...
-            const updatedDate = feature.getProperty('updated_at');
-            if (updatedDate < breakpointDate) {
+            const plantedDate = feature.getProperty('plantingDate');
+            const adopted = !!feature.getProperty('adopterId');
+            if (adopted) {
+              icon = adoptedTreeIcon;
+            } else if (!!plantedDate && plantedDate > breakpointDate) {
```
SofieDunt, line 383: ❓ What's the difference between putting `!!` here versus before `feature.getProperty('plantingDate')`?

SofieDunt (commented):
Looks solid, just left a couple questions

src/components/mapPageComponents/mapView/index.tsx
willmt80: This notation (!!) casts the returned value to a boolean based on whether or not it is truthy. Because we need to know both whether or not a planting date exists and what the value of that date is, we have to assign the actual value to a variable first

src/components/mapPageComponents/ducks/types.ts
willmt80: Yes, I believe these have to be optional because the api may return null for these values. Good catch on the address, that should be optional as well

SofieDunt (approved):
💯

PR #76, opened 2021-05-26
added my trees page and moved mapping state to props to geojson to container level

willmt80:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377)

Users can now look at a list of the trees they have adopted alongside a tree map. Later on we can add the option to display only a users adopted trees on the map.

## This PR

I based most of the myTrees page on the reservation page. The new components for cards and for the map page use modified versions of components that already exist. There is plenty of opportunity here for abstraction but seeing as the demo is *tomorrow* getting the working page up was the highest priority. 

In order to have access to the sites for the sidebar, I decided to move mapping the geojson to state to the container level. This means that mapContent now takes in the requests as props from the parent. Ideally this will make it much easier to integrate stuff like finding nearby trees and blocks as the the sidebars and maps will be looking at the same data.

The page can be accessed via the "My Trees" link on the home page.

## Screenshots

 

my-trees

my-trees-mobile

**EDIT: THE DUPLICATE ID IS FROM MY LOCAL BACKEND, NOT AN ISSUE WITH THE SELECTOR**

## Verification Steps

 

I tested that the sidebar can display the tree information correctly and I made sure that the buttons take the user to the tree page associated with the correct ID. Sometimes I got a bug that causes an error on the whole page but that is a problem with the backend route.


src/assets/content.tsx
```diff
@@ -34,6 +34,11 @@ export const RESERVATION_TITLE = 'My Blocks';
 export const RESERVATION_BODY =
   "Here are the blocks you can sign up to inventory and the ones you've reserved! Did you know that we take care of over 500 blocks of trees? We couldn't do it without you! ";
 
+// My TRees
+export const MY_TREES_TITLE = 'My Trees';
+export const MY_TREES_BODY =
+  'Here are the trees you have adopted. From here you can select a tree to see more information or you can adopt even more!';
```
SofieDunt, line 40: ⛏️ Maybe mention recording stewardship activities instead of seeing more information

src/components/treeCard/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+    <>
+      <StyledCard>
+        <CardContent>
+          <BlockInfo>
+            {site.address && <TreeTitle>Address: {site.address}</TreeTitle>}
+            {site.species && <TreeBody>Species: {site.species}</TreeBody>}
+            {site.id && <TreeBody>ID: {site.id}</TreeBody>}
+          </BlockInfo>
+          <GreenLinkButton to={`${ParameterizedRouteBases.TREE}${site.id}`}>
+            Tree Page
```
SofieDunt, line 65: ⛏️ Maybe `Record Stewardship` or `More Info`

src/components/treeSidebar/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+const StyledListItem = styled(List.Item)`
+  padding: 3px;
+`;
+
+interface TreeSidebarProps {
+  readonly mySites: SiteFeaturePropertiesResponse[];
+}
+
+const BlockTabs: React.FC<TreeSidebarProps> = ({ mySites }) => {
```
SofieDunt, line 20: ❓ `const TreeSidebar`? Worth looking into abstracting the two

src/containers/myTrees/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+const MyTrees: React.FC<MyTreesStateProps> = ({
+  blocks,
+  neighborhoods,
+  sites,
+}) => {
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    dispatch(getMapGeoData());
+    dispatch(getAdoptedSites());
```
SofieDunt, line 37: 📌 ❓ Do we dispatch to get the map data every time this page/landing/reservations renders? If so, I think we should consider only dispatching if the state isn't complete/hasn't been updated in a while. Not too sure how this works and not in the scope of this ticket but I've noticed it seems to get slower on subsequent reloads

SofieDunt (commented):
Pictures look great! We should take the time to clean everything up now that demo day is over, sorry for rushing that

src/containers/myTrees/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+            return (
+              <MapPage
+                sidebarHeader={MY_TREES_TITLE}
+                sidebarDescription={MY_TREES_BODY}
+                view={treeMapView}
+                blocks={blocks}
+                neighborhoods={neighborhoods}
+                sites={sites}
+              >
+                <TreeSidebar mySites={mySites} />
```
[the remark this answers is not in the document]
willmt80: Changed this to a message that says "No Trees Adopted"

src/containers/myTrees/index.tsx
willmt80: Yeah Ryan and I had a conversation about this yesterday. We're thinking of putting in a timer so that the calls are more staggered

src/assets/content.tsx
```diff
@@ -55,6 +55,11 @@ export const RESERVATION_TITLE = 'My Blocks';
 export const RESERVATION_BODY =
   "Here are the blocks you can sign up to inventory and the ones you've reserved! Did you know that we take care of over 500 blocks of trees? We couldn't do it without you! ";
 
+// My TRees
```
SofieDunt, line 58: ⛏️ `Trees`

src/components/blockCard/index.tsx
```diff
@@ -45,9 +41,9 @@ const BlockCard: React.FC<BlockCardProps> = ({ id, reserved }) => {
     <>
       <StyledCard>
         <CardContent>
-          <BlockInfo>
+          <CardInfo>
             <BlockTitle>Block {id}</BlockTitle>
-          </BlockInfo>
+          </CardInfo>
```
SofieDunt, line 46: 💯

src/components/blockTabs/index.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { List, Tabs } from 'antd';
 import styled from 'styled-components';
+import { StyledListItem, ScrollableListContainer } from '../themedComponents';
```
SofieDunt, line 4: 💯

src/components/treeCard/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  readonly site: SiteFeaturePropertiesResponse;
+}
+
+const TreeCard: React.FC<TreeCardProps> = ({ site }) => {
+  return (
+    <>
+      <StyledCard>
+        <CardContent>
+          <CardInfo>
+            {site.address && <TreeTitle>Address: {site.address}</TreeTitle>}
```
SofieDunt, line 56: Every site should have a zip code; maybe put that if there's no address.

src/components/treeSidebar/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+const TreeSidebar: React.FC<TreeSidebarProps> = ({ mySites }) => {
+  return (
+    <>
+      <ScrollableListContainer>
+        <List
+          dataSource={mySites}
+          itemLayout="vertical"
+          locale={{ emptyText: 'No Trees Adopted' }}
```
SofieDunt, line 18: 👍

SofieDunt (approved):
🎉 Looks great!

src/components/treeCard/index.tsx
willmt80: Eventually yes. Right now the zip is not returned as a part of the site feature so that will require another ticket

PR #78, opened 2021-06-08
Add Planting Sites to Map and Format Tree Data

SofieDunt:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1369284383)

Makes adjustments based on SFTT's feedback.

## This PR

Major Changes:
1. Tree icons resize on zoom
2. Planting sites are visible and have a separate icon
3. Some tree data is initially hidden
4. NavBar structure - using AntD's pageheader component made styling difficult after adding the partner logos to the title, no longer uses this

The rest of the changes are mostly to fix wording.

Cleanup:
1. Removed some catch statements so that requests are dispatched as failed on errors.

## Screenshots

 
image
image
image
image
image
UPDATE: NEW NAV BAR AND ENTRY ORDER
image


src/assets/content.tsx
```diff
@@ -9,41 +9,51 @@ export const HOME_HEADER = "Let's get back to those trees, why don't we?";
...
 export const ICONS_DIRECTION =
-  'Click on the icons of trees you might be interested in adopting.';
+  'Click on the dot of a tree you might be interested in adopting.';
 export const REDIRECTED_DIRECTION =
   'Once you\'ve found a tree to adopt, click on "More Info" to go to that tree\'s page, where you can adopt it!';
 
 // Map Legend
 export const ADOPTED_TREE_ICON_DESCRIPTION =
-  "can't be adopted - they already have someone looking after them!";
+  'These trees have already been adopted';
```
willmt80, line 28: ⛏️  "Trees that have already been adopted" is more consistent with the language used for the rest of the key

src/api/protectedApiClient.ts
```diff
@@ -226,9 +226,9 @@ const changePrivilegeLevel = (
 };
 
 const getUserData = (): Promise<UserData> => {
-  return AppAxiosInstance.get(ProtectedApiClientRoutes.GET_USER_DATA)
-    .then((res) => res.data)
-    .catch((err) => err);
+  return AppAxiosInstance.get(ProtectedApiClientRoutes.GET_USER_DATA).then(
+    (res) => res.data,
```
willmt80, line 230: 📌 Not for this ticket, it seems like we want to remove catching from every one of these route calls so we can add some error condition where they are called.

src/containers/treePage/index.tsx
```diff
@@ -185,104 +175,106 @@ const TreePage: React.FC<TreeProps> = ({ siteData, stewardship, tokens }) => {
...
-            />
-          </Row>
+                        <MobileTreeCareContainer>
+                          <TreeActivity stewardship={stewardship} limit={4} />
+                        </MobileTreeCareContainer>
+                      </MobileTreeMainContainer>
+                    );
+                }
+              })()}
+              <CenterDiv>
```
willmt80, line 255: 💯  I like the way you split these up

src/utils/stringFormat.tsx
```diff
@@ -48,3 +50,63 @@ export function shortHand(
...
+
+/**
+ * If both are present in the given list of entries, combines genus and species into a scientific name.
+ * @param entries the list of entries
+ */
+export function combineScientificName(entries: Entry[]): Entry[] {
+  const newEntries: Entry[] = [];
+  let species;
+  let genus;
+  entries.forEach((entry: Entry) => {
```
willmt80, line 90: 💭 There are a few places where we seem to be wrestling between arrays and records for site entries, it might be easier if transform the list into a record or map in the selector so that way we can do everything by name after

src/utils/test/stringFormat.test.ts
```diff
@@ -1,7 +1,14 @@
-import { formatDateSuffix, getMoneyString } from '../stringFormat';
+import {
```
willmt80, line 1: 🔥

willmt80 (approved):
Look good! Left a few thoughts and changes. Again, not super ideal because we need to deploy soon and get some changes deployed so we can put a pin in some of the more significant changes

src/components/navBar/index.tsx
```diff
@@ -79,10 +128,32 @@ const NavBar: React.FC<NavBarProps> = ({ tokens, userData }) => {
...
-      <Title level={3}>Speak for the Trees</Title>
-    </Button>
+  const HeaderTitle = () => (
+    <NoHoverShadeButton type="text" onClick={() => history.push(Routes.HOME)}>
+      <Row align="bottom">
+        <Col span={6}>
+          <MainLogo src={sfttLogo} alt="icon" />
+        </Col>
+        <Col span={6}>
+          <NavTitleText>in partnership with</NavTitleText>
```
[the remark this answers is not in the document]
SofieDunt: This was from SFTT but I agree, we'll go with this for now and get Brandon to take a look in the fall.

src/assets/content.tsx
SofieDunt: This was from SFTT

src/api/protectedApiClient.ts
SofieDunt: https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1375952699

src/utils/stringFormat.tsx
SofieDunt: Will look into this in a separate ticket

PR #81, opened 2021-06-17
Site geo json query update

willmt80:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1398350639)

No longer uses the scientific species name for a tree, instead uses the common name. Also removes the fields that have been removed from the map site query in this PR https://github.com/Code-4-Community/speak-for-the-trees-backend-v2/pull/61

## This PR

Literally just removed fields and changed 1 name in 1 object and changed the property fetched for the tree popup

## Screenshots

 

Capture

Notice all the new blue sites

## Verification Steps

 

I checked some trees popups to make sure their names were correct. Also for verification of the backend PR, I can see all of the new blue sites from inserting data earlier


src/components/mapPageComponents/ducks/types.ts
```diff
@@ -63,15 +63,12 @@ interface SiteFeatureResponse {
 export interface SiteFeaturePropertiesResponse {
   id: number;
   treePresent: boolean;
-  diameter?: number;
-  species?: string;
-  updatedAt: string;
-  updatedBy: string;
-  address?: string;
```
SofieDunt, line 70: 💯

src/components/mapPageComponents/mapView/index.tsx
```diff
@@ -378,7 +378,7 @@ const MapView: React.FC<MapViewProps> = ({
           // Sets the information to display in the popup
           setActiveTreeInfo({
             id: siteId,
-            species: eventFeature.getProperty('species'),
+            species: eventFeature.getProperty('commonName'),
```
SofieDunt, line 381: Do you know why we use camelCase here and snake_case elsewhere (like `tree_present` and `block_id`)?

SofieDunt (approved):
Looks good!

src/components/mapPageComponents/mapView/index.tsx
willmt80: There's no good reason for us to use snake case anywhere in the frontend

PR #94, opened 2021-06-23
changed navbar alignment and popup dimensions

willmt80:
## Why

[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1401865519)
[Monday.com Ticket](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1398345228)

Formatting fixes to map popups on the mapView look nicer. Removes the unnecessary horizontal and vertical scrolling. Also changed the alignment of the nav bar logos so they do not get cut off when a screen is short

## This PR

I added a min-height to the tree popups to set their size and I removed the max height to prevent text from being cut off. The common name and address already have length limits enforced in the database so there is not a concern about a huge chunk of text in the popup. 

For the navbar I changed the row alignment from bottom to middle and set a min height for the navbar so if the user makes their window short the navbar won't cut off the logos. I also made the padding for the login and signup buttons a constant to prevent them from moving towards the top of the screen as the window gets shorter.

## Screenshots

 

format

## Verification Steps

 

I changed the size of the screen both vertically and horizontally and I checked on many trees to ensure the popup did not have weird proportions

### DIFF
--- a/src/components/navBar/index.tsx
+++ b/src/components/navBar/index.tsx
@@ -54,10 +54,12 @@ const MainLogo = styled.img`
 
 const BostonLogo = styled.img`
   height: 55px;
+  line-height: 0px;
 `;
 
 const C4CLogo = styled.img`
   height: 27px;
+  line-height: 0px;
 `;
 
 const LandingExtraContainer = styled.div`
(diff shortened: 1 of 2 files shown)


SofieDunt (approved):
💯

PR #108, opened 2021-09-19
Improve Tree Page-Map Return

SofieDunt:
## Why

[ClickUp Ticket](https://app.clickup.com/t/1dk42tb)

Users are having difficulty switching between the map and tree pages. Improves user flow by opening the tree page in a new tab and opening the map at the street zoom level at the site location via the 'Return to Tree Map' button.

## This PR

- use Link target to open tree page in a new tab
- use location state to pass the map the site's location and the zoom level to go to
  - ensure layer visibility is initialized according to the zoom level rather than assuming the map is zoomed out
- clean up codebase by putting the redirect button under themed components

## Verification Steps

 
Tested going from the map to the tree page (check that page opens in a new tab) and returning to the map (check that the map is zoomed in at the site's location)
Tested going from a tree page to the map via the return button (check that the map is zoomed in at the site's location)


willmt80:
It looks like coveralls is down so hopefully this build will pass once it's back up.

src/components/mapPageComponents/logic/init.ts
```diff
@@ -40,12 +44,14 @@ export function initPrivateStreets(map: google.maps.Map): google.maps.Data {
...
  * @param view the map view
  * @param map the map to add the layer to
+ *  @param visible whether to initiate the layer as visible
  */
 export function initNeighborhoods(
   neighborhoods: NeighborhoodGeoData,
   markersArray: google.maps.Marker[],
   view: MapViews,
   map: google.maps.Map,
+  visible: boolean,
```
willmt80, line 54: 👍

src/components/mapPageComponents/mapView/index.tsx
```diff
@@ -124,11 +128,15 @@ const MapView: React.FC<MapViewProps> = ({
...
     if (mapElement && treePopupElement) {
+      const defaultZoom: number = location.state ? location.state.zoom : 12;
+
       LOADER.load()
         .then(() => {
           map = new google.maps.Map(mapElement, {
-            center: BOSTON,
-            zoom: 12,
+            center: location.state
+              ? { lat: location.state.lat, lng: location.state.lng }
```
willmt80, line 137: Not too familiar with useLocation. Is it possible to set the initial state to BOSTON so we can avoid this ternary operator?

src/components/themedComponents/index.tsx
```diff
@@ -53,6 +54,13 @@ export const SubmitButton = styled(Button)`
   font-size: 16px;
 `;
 
+export const ReturnButton = styled(LinkButton)`
```
willmt80, line 57: 👍

src/containers/treePage/index.tsx
```diff
@@ -33,14 +33,21 @@ import { RecordStewardshipRequest } from '../../components/forms/ducks/types';
...
-import ReturnButton from '../../components/returnButton';
 import TreeInfo from '../../components/treeInfo';
 import TreeActivity from '../../components/treeActivity';
 import EntryList from '../../components/entryList';
 import { CenterDiv } from '../../components/themedComponents';
+import { STREET_ZOOM } from '../../components/mapPageComponents/constants';
 
 const { Title } = Typography;
 
+const ReturnButton = styled(Button)`
```
willmt80, line 44: Can we give the link button an optional state argument? that way we don't need to defined this style or link functionality below more than once. Plus I'm sure we'll wind up reusing this for the blocks in the future.

willmt80 (changes requested):
Left a couple of suggestions/questions, other than that looks good!

adriannaliu (approved):

src/components/linkButton/index.tsx
```diff
@@ -1,18 +1,13 @@
...
-
+export const LinkButton: React.FC<any> = ({ to, state, children, ...rest }) => {
   return (
-    <Button
-      {...rest}
-      onClick={() => {
-        history.push(to);
-      }}
-    />
+    <Button {...rest}>
```
willmt80, line 9: 🔥

willmt80 (approved):
Looks good!

PR #110, opened 2021-09-21
Remove Private Streets From Map

adriannaliu:
## Why

[ClickUp Ticket](https://app.clickup.com/t/1ddkhz5)

Private street layers' visibility is initially set to false but when the map zooms in, the private streets (red lines) become visible. The specific feature for the private street layer will not supported in the near future, so it's currently unnecessary for the layer to be visible in the map view.

## This PR

in addHandleZoomChange, simply set the boolean parameter for setPrivateStreetsStyle to false


src/components/mapPageComponents/logic/event.ts
```diff
@@ -131,7 +131,7 @@ export function addTreePopupOnClick(
  * Adds an event listener to handle zoom changes.
  * @param neighborhoodsLayer the neighborhoods layer
  * @param markersArray the array of neighborhood label markers
- * @param privateStreetsLayer thee private streets layer
+ * @param privateStreetsLayer the private streets layer
```
SofieDunt, line 134: 👍

src/components/mapPageComponents/logic/event.ts
```diff
@@ -156,7 +156,7 @@ export function addHandleZoomChange(
       zoomedIn = true;
     }
     setNeighborhoodsStyle(neighborhoodsLayer, markersArray, !zoomedIn);
-    setPrivateStreetsStyle(privateStreetsLayer, zoomedIn);
+    setPrivateStreetsStyle(privateStreetsLayer, !zoomedIn);
```
SofieDunt, line 159: What this will do is make it so that when a user zooms in, the streets aren't visible, which is good but it'll also make the streets visible when you zoom out. We want to remove visibility at all zoom levels. Changing `!zoomedIn` to `false` would do the trick, or you could also mess with the MapView file directly so that the streets never get loaded into the map at all (if you do this, you'd also have to change this function and anywhere else that sets the private streets style)

SofieDunt (changes requested):
🔥 Just need a minor adjustment for an edge case!

src/components/mapPageComponents/logic/event.ts
adriannaliu: got it! I think I still want to keep the functionality to have the option to load the streets on the map but this is good to know

SofieDunt (approved):
👍  Looks good to me!

vrushalitarte (approved):
looks good! congrats on your first PR!

PR #111, opened 2021-09-23
Displays information on tree page if common name is unknown

lnardini:
## Why

[ClickUp Ticket](https://app.clickup.com/t/1fqtbjh)

Allows tree pages to have headers if they have no common name or species.  

## This PR

Changed logic for displaying header.  Formerly header was dependent on common name existing, replaced with ? operator. 

## Screenshots

image

## Verification Steps

Compared https://map.treeboston.org/tree/3215900 and http://localhost:3000/tree/3215900: former has a blank header where local version contains address information

### DIFF
--- a/src/components/treeInfo/index.tsx
+++ b/src/components/treeInfo/index.tsx
@@ -68,14 +68,19 @@ const TreeInfo: React.FC<TreeProps> = ({
   return (
     <>
       <TreeHeader>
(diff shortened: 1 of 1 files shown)


SofieDunt (approved):
👍  Looks good to me!

chromium-52 (approved):
Looks great!

PR #116, opened 2021-10-22
Show stewardship activities per month

chromium-52:
## Why

[ClickUp Ticket](https://app.clickup.com/t/1dk45bd)

On the tree page, remove the limit on the number of stewardship activities shown and show stewardship activities in a specified month, where the default month is the current month.

## This PR

Added a dropdown menu that the user can use to show stewardship activities in the selected month.
Added a pagination component below the stewardship activity list (which appears only if there are more than 10 stewardship activities that were recorded during the month selected by the user) to reduce the amount of scrolling.

## Screenshots

At most 10 activities shown:
first

More than 10 activities shown:
second

 

## Verification Steps

 
Visual verification that the planting instruction message displayed is correct on Firefox. Ran `npm run check` and `npm run test`.


src/components/treeActivity/index.tsx
```diff
@@ -35,19 +35,146 @@ const EntryMessage = styled(Paragraph)`
...
+        ),
+    )
+    // turn the filtered array into an array of labels and values
+    .map((entry) => {
+      return {
+        label: monthShortToLong(entry.date) + ' ' + entry.year,
+        value: entry.date.substring(0, 3) + ' ' + entry.year,
+      };
+    });
+  // append the current month and year to the options list
```
SofieDunt, line 140: ❓ What happens if one of the stewardship activities took place in the current month and year?

src/components/treeActivity/index.tsx
```diff
@@ -35,19 +35,146 @@ const EntryMessage = styled(Paragraph)`
...
+`;
+
 interface TreeActivityProps {
   readonly stewardship: TreeCare[];
-  readonly limit?: number;
 }
 
-const TreeActivity: React.FC<TreeActivityProps> = ({ stewardship, limit }) => {
+const TreeActivity: React.FC<TreeActivityProps> = ({ stewardship }) => {
+  const months = [
```
SofieDunt, line 63: ⛏️ I think it'd be better to store the abbreviated and un-abbreviated months in a dictionary (see `SHORT_HAND_NAMES` under `src/assets/content`). That way you can also use the `shortHand` function under `src/utils/stringFormat`, which should be more efficient than `monthShortToLong`.

src/components/treeActivity/index.tsx
```diff
@@ -35,19 +35,146 @@ const EntryMessage = styled(Paragraph)`
...
+    for (const month of months) {
+      if (shortMonth.substring(0, 3) === month.substring(0, 3)) {
+        return month;
+      }
+    }
+    return '';
+  };
+
+  // sort two dates by a combination of year and month
+  const sortByYearMonth = (
```
SofieDunt, line 89: 💯 Can you extract this and put it under a new file in `utils` so we can also improve our test coverage?

src/components/treeActivity/index.tsx
```diff
@@ -35,19 +35,146 @@ const EntryMessage = styled(Paragraph)`
...
+          (e) =>
+            e.year === entry.year &&
+            e.date.substring(0, 3) === entry.date.substring(0, 3),
+        ),
+    )
+    // turn the filtered array into an array of labels and values
+    .map((entry) => {
+      return {
+        label: monthShortToLong(entry.date) + ' ' + entry.year,
+        value: entry.date.substring(0, 3) + ' ' + entry.year,
```
SofieDunt, line 137: 💭 You could have `value` be of type `{month: string, year: number}` instead of just `string` to avoid having to extract the month vs year everywhere with `substring` and `splice`

src/containers/treePage/ducks/types.ts
```diff
@@ -117,6 +117,7 @@ export const ExtraSiteEntryNames: Record<string, string> = {
 };
 
 export interface TreeCare {
+  year: number;
```
SofieDunt, line 120: 👍

src/containers/treePage/test/selectors.test.ts
```diff
@@ -50,10 +50,12 @@ describe('Tree Page Selectors', () => {
       const expectedTreeCare: TreeCare[] = [
         {
           date: 'Jan 1st',
+          year: 2021,
           message: 'Was mulched and watered and weeded.',
         },
         {
           date: 'Feb 23rd',
+          year: 2021,
```
SofieDunt, line 58: 👍

SofieDunt (commented):
Looks good! Left a ❓ and a ⛏️

src/components/treeActivity/index.tsx
willmt80: ☝️

src/components/treeActivity/index.tsx
```diff
@@ -35,19 +35,146 @@ const EntryMessage = styled(Paragraph)`
...
+      year: 'numeric',
+    }),
+  );
+  const selectedMonthYearStewardship: TreeCare[] = stewardship.filter(
+    (entry) =>
+      // choose stewardship activities with the selected month and year
+      entry.date.substring(0, 3) === selectedMonthYear.substring(0, 3) &&
+      entry.year === parseInt(selectedMonthYear.slice(-4), 10),
+  );
+  const monthYearOptions = stewardship
```
willmt80, line 122: It makes sense to move a lot of this functionality to a selector. Instead of getting the formatted TreeCare from the store via the mapStewardshipToTreeCare and then filtering and recreating the list based on a parsed string, you can use the date values stored in Redux.

willmt80: You also want child components to be kind of stupid, it's better to do a lot of this computation up top in the parents and then just leave the children to display what they are given

src/components/treeActivity/index.tsx
```diff
@@ -35,19 +35,146 @@ const EntryMessage = styled(Paragraph)`
...
+    if (year1 === year2 && month1 === month2) {
+      return 0;
+    } else if (year1 < year2 || (year1 === year2 && month1 < month2)) {
+      return -1;
+    } else {
+      return 1;
+    }
+  };
+
+  const [selectedMonthYear, setSelectedMonthYear] = useState(
```
willmt80, line 110: It's generally better to have state values like this in parent components and then pass the set methods as props

willmt80 (changes requested):
💯 Nice work on the design and the pagination

Generally I think we want to move a lot of this functionality out of the TreeActivity component and into utils files, the parent component, and a selector.

src/components/treeActivity/index.tsx
SofieDunt: ☝️

src/components/treeActivity/index.tsx
SofieDunt: @willmt80 I think it's better to have this state val in the child in this case - the `selectedMonthYear` is only relevant to the child component, which I believe means it would be better to only re-render the component vs the entire page when updating the val, as it would if it was in the parent?

src/components/treeActivity/index.tsx
chromium-52: Oh interesting. I think I fixed this by flipping the orders in which the current month/year was added to `monthYearOptions` and the duplicate elements in `monthYearOptions` were removed.

src/components/treeActivity/index.tsx
chromium-52: I tried doing this but couldn't because the `options` attribute of the `Select` component wants something of type `(OptionData | OptionGroupData)[]` and `{ label: string, value: { month: string, year: number } }` doesn't match that type. Specifically, it's not an `OptionData` because the `value` field is not of type `Key` (which is an alias for `string | number`). But it's also not an `OptionGroupData` because it doesn't have the `options` field. Should I leave the `value` field as is and add a third field of type `{ month: string, year: number }` or is there a better way to do it?

image

image

src/components/treeActivity/index.tsx
chromium-52: Should I put `selectedMonthYear` in the parent or child component?

willmt80: Yeah Sofie is right, leaving this in the child is good

src/components/treeActivity/index.tsx
```diff
@@ -35,19 +38,72 @@ const EntryMessage = styled(Paragraph)`
...
+  monthYearOptions,
+}) => {
+  const [selectedMonthYear, setSelectedMonthYear] = useState(
+    new Date().toLocaleString('default', {
+      month: 'short',
+      year: 'numeric',
+    }),
+  );
+  const selectedMonthYearStewardship: TreeCare[] = stewardship.filter(
+    (entry) =>
```
willmt80, line 77: We should change the way TreeCare/entries are stored. It doesn't make much sense to convert the stewardship to  strings then do a bunch of substring stuff in the filtering. Change TreeCare to have month, day, and year in the exact format you need to filter them then convert only the filtered values to the correct string formats after.

willmt80 (changes requested):
Moving stuff to the selectors looks good 👍. I think we should do some more dramatic changes to the TreeCare organization, that type was not originally created with filtering in mind. Let me know if you have any questions and we can hop on a call.

src/components/treeActivity/index.tsx
```diff
@@ -35,19 +38,72 @@ const EntryMessage = styled(Paragraph)`
...
+  );
+
+  const [pageNumber, setPageNumber] = useState(0);
+
   return (
     <>
       <TreeCareTitle>Recent Tree Care Activity</TreeCareTitle>
+      <StewardshipActivityDropdownContainer>
+        <StewardshipActivityDropdown>
+          Month and year to display activities for:{' '}
```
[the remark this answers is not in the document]
chromium-52: It was to add a space between the text and the dropdown menu. Is there a different way to do this?

src/containers/treePage/ducks/selectors.ts
```diff
@@ -85,3 +91,40 @@ export const isTreeAdoptedByUser = (
...
+            shortHand(entry.date.substring(0, 3), UNABBREVIATED_MONTHS) +
+            ' ' +
+            entry.year,
+          value: entry.date.substring(0, 3) + ' ' + entry.year,
+        };
+      })
+      // append the current month and year to the options list
+      .concat([
+        {
+          label: new Date().toLocaleString('default', {
```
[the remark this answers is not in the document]
chromium-52: Fixed! :)

rymaju (approved):
LGTM

src/containers/treePage/ducks/selectors.ts
```diff
@@ -85,3 +91,34 @@ export const isTreeAdoptedByUser = (
...
+      // remove entries with duplicate month and year
+      .filter(
+        (entry, index, self) =>
+          index ===
+          self.findIndex(
+            (e) => e.month === entry.month && e.year === entry.year,
+          ),
+      );
+    // sort dates by both month and year in reverse chronological order
+    return monthYearOptions.sort(compareByMonthYear).reverse();
```
SofieDunt, line 122: 💭  You could have compareByMonthYear return reverse chronological order instead of reversing after, but the list will be small so it shouldn't matter.

SofieDunt (approved):
Great work! 💯

PR #118, opened 2021-11-13
Edit Site Page

SofieDunt:
## Why

[ClickUp Ticket](https://app.clickup.com/t/1rp4qhc)

Creates a page to let admins edit features of sites + update site information through the frontend rather than by sending us the data.

Makes several updates to Google Maps API logic and components that will make implementing additional maps simpler and less repetitive in the future.

*UPDATE - Fixes types and creates enum for site entry fields to make future changes less prone to error.*

*Creates several themed components to make flexible layouts easier to construct in the future.*

## This PR

This ticket is for desktop designs only. *UPDATE - layout with flex works reasonably well on different window sizes*

Map Restructuring Logic:
- mapComponents contains all components related to the maps
- maps contains all actual maps, all current maps are based on MapWithPopup, which keeps the logic for loading the map, creating the popup, and executing search functions in the same place. It takes an initMap function which sets up features unique to each map, such as which data layers to load and whether or not the search marker can be dragged
- mapDisplays contains components that render the map only if the map data has loaded, otherwise it shows the loading icon or an error message accordingly. *Containers/Parent components of maps should use displays, not the map itself*
- mapPageComponents contains components concerned with layout, not map logic

SelectorMap:
- dragging the search marker will update the form's lat/lng, but not vice-versa - I believe the change would require the map to re-render on every form update, which is not ideal

## Screenshots - *UPDATED WITH FLEX LAYOUT*

 

image

image

image

image

## Verification Steps

 
Tested that the forms work, added appropriate redux/function tests, interacted with maps on the frontend to ensure old components also worked on landing/my trees page


src/App.tsx
```diff
@@ -102,6 +105,17 @@ const App: React.FC = () => {
...
+                        path={Routes.SITE}
+                        exact
+                        render={({ match }) => {
+                          return (
+                            <AuthRedirect
+                              from={`${ParameterizedRouteBases.SITE}${match.params.id}`}
+                            />
+                          );
+                        }}
+                      />
```
SofieDunt, line 118: This is to pass the path params to the redirect as well, otherwise the redirect would go back to `/:id`, which would cause an error

src/assets/content.tsx
```diff
@@ -94,3 +94,90 @@ export const EMISSIONS_STAT_TITLE = 'Carbon Emissions Captured';
...
+  { label: 'Chinatown', value: 26 },
+  { label: 'Leather District', value: 27 },
+  { label: 'Longwood', value: 28 },
+  { label: 'South Boston Waterfront', value: 29 },
+  { label: 'Beacon Hill', value: 30 },
+  { label: 'West End', value: 31 },
+  { label: 'South End', value: 32 },
+  { label: 'Bay Village', value: 33 },
+  { label: 'Fenway', value: 34 },
+];
```
SofieDunt, line 183: Painfully redundant but quicker than going through the label/value pairs when trying to translate an id to a name - any other suggestions?

src/containers/sitePage/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+const SitePage: React.FC<SitePageProps> = ({ neighborhoods, sites }) => {
+  const [site, setSite] = useState<SiteProps>();
+  const id = Number(useParams<SiteParams>().id);
+  const { windowType } = useWindowDimensions();
+
+  const [editSiteForm] = Form.useForm();
+  const [updateSiteForm] = Form.useForm();
+
+  const getSite = () => {
+    // TODO: BUG - getSite isn't getting neighborhood id
```
SofieDunt, line 68: Did some further investigation and it turns out the backend returns the neighborhood name instead of ID. We can either update the docs + this frontend code or the backend. I'd prefer the backend, but open to any reasons for keeping the backend as is. Currently, this discrepancy just means the edit site page will always say 'No neighborhood found.' instead of the neighborhood name, which isn't ideal but doesn't break anything

src/utils/formRules.tsx
```diff
@@ -91,3 +91,7 @@ export const newLevelRules: Rule[] = [
     message: 'Please pick a privilege level for this user!',
   },
 ];
+
+export const requiredRule = (message: string): Rule[] => {
+  return [{ required: true, message }];
+};
```
SofieDunt, line 97: 📌 We should update the other forms to use this in a separate PR

src/containers/sitePage/index.tsx
willmt80: This is a quick change

src/components/mapComponents/ducks/types.ts
```diff
@@ -103,3 +104,29 @@ export enum MapViews {
   BLOCKS = 13,
   TREES = 16,
 }
+
+export interface BasicMapData {
```
willmt80, line 108: 🔥

willmt80: Nice refactoring

src/components/mapComponents/mapDisplays/treeMapDisplay/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+`;
+
+interface TreeMapDisplayProps {
+  readonly neighborhoods: MapGeoDataReducerState['neighborhoodGeoData'];
+  readonly sites: MapGeoDataReducerState['siteGeoData'];
+  readonly mobile: boolean;
+  readonly returnTo?: Routes;
+}
+
+const TreeMapDisplay: React.FC<TreeMapDisplayProps> = ({
```
willmt80, line 25: 👍

src/components/mapComponents/mapPageComponents/mapPage/index.tsx
```diff
@@ -1,43 +1,32 @@
...
-import { MapGeoDataReducerState, MapViews } from '../ducks/types';
-import MapLegend from '../mapLegend';
-import { WindowTypes } from '../../windowDimensions';
+import { MainContent } from '../../../themedComponents';
+import { MapViews } from '../../ducks/types';
+import MapLegend from '../../mapLegend';
+import { WindowTypes } from '../../../windowDimensions';
 
 interface MapPageProps {
+  readonly mapContent: JSX.Element;
```
willmt80, line 11: ❓ Does this give you any console errors? I know I've had some trouble passing JSX as props like this

src/components/mapComponents/maps/blocksMap/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+      searchMarker,
+      zoomListener: mapLayersAndListeners.zoomListener,
+      markersArray: mapData.markersArray,
+    };
+  };
+
+  return (
+    <MapWithPopup
+      zoom={defaultZoom}
+      view={MapViews.TREES}
```
willmt80, line 58: Should be BLOCKS?

src/components/mapComponents/maps/blocksMap/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+      lat={defaultCenter.lat}
+      lng={defaultCenter.lng}
+      initMap={initMap}
+      defaultActiveTree={{
```
willmt80, line 62: 📌  Will need to have active blocks, not trees here in the future

src/components/mapComponents/maps/treeMap/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+      markersArray: mapData.markersArray,
+    };
+
+    setLoadedMapData(setMapData);
+
+    return setMapData;
+  };
+
+  // Add new zoom listener and update sites style whenever visibleSites changes
+  const onCheck = (values: CheckboxValueType[]): void => {
```
willmt80, line 68: ❓  Will this zoom functionality be dramatically different for blocks? it seems like the two map pages will have almost identical code with just a couple of different parameters/props.

src/components/themedComponents/index.tsx
```diff
@@ -60,6 +60,23 @@ export const SubmitButton = styled(Button)`
   font-size: 16px;
 `;
 
+export const GreenButton = styled(Button)`
```
willmt80, line 63: Should this component just be used everywhere? Like for login and on all form submission?

src/containers/landing/index.tsx
```diff
@@ -105,7 +112,6 @@ const Landing: React.FC<LandingProps> = ({ blocks, neighborhoods, sites }) => {
 const mapStateToProps = (state: C4CState): LandingProps => {
   return {
     neighborhoods: state.mapGeoDataState.neighborhoodGeoData,
-    blocks: state.mapGeoDataState.blockGeoData,
```
willmt80, line 108: 👍

willmt80 (changes requested):
💯 Great work on this! Left a few questions and small comments about performance and refactoring

src/assets/content.tsx
SofieDunt: Maybe a getNeighborhoodNames route that we can save in store? I think IDs/names are translated too often to rely on a getNeighborhoodName route. We can keep these for now and update in a future ticket - lmk what kind of backend route you think is best.

src/containers/sitePage/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+const MarginBottomRow = styled(Row)`
+  margin-bottom: 30px;
+`;
+
+interface SiteParams {
+  id: string;
+}
+
+interface SitePageProps {
+  readonly neighborhoods: MapGeoDataReducerState['neighborhoodGeoData'];
```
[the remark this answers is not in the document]
SofieDunt: It's not terrible for me. The map component renders independently from the rest of the page, so I think it's fine to have the rest of the page loaded while it takes a bit longer to fetch the data/render the map.

src/components/forms/updateSiteForm/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+      style={{ width: '100%' }}
+    >
+      <Row gutter={[16, 8]}>
+        <TitleCol span={6}>Is there a tree present?</TitleCol>
+        <TitleCol span={6}>Is the tree multi-stem?</TitleCol>
+        <TitleCol span={6}>Is the tree discolored?</TitleCol>
+        <TitleCol span={6}>Is the tree leaning?</TitleCol>
+      </Row>
+      <Row gutter={[16, 8]}>
+        <BoolRadioCol name={'treePresent'} />
```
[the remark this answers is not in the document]
SofieDunt: Set these to false with `initialValues` in the form vs `defaultValue` in the item because the latter gave a console warning.

src/components/mapComponents/mapPageComponents/mapPage/index.tsx
SofieDunt: I'm not seeing any–what kind of errors were they?

src/components/mapComponents/maps/treeMap/index.tsx
SofieDunt: Yeah, most of the zoom logic is abstracted to `addHandleZoomChange` (which this function uses). This function is unique because when we hide some sites, we have to remove and re-add the zoom handler, then turn on the site styling if we're already zoomed in (which the handler doesn't initially check for).

src/components/themedComponents/index.tsx
SofieDunt: Probably - https://app.clickup.com/t/200h9q5

src/components/forms/ducks/types.ts
```diff
@@ -36,3 +41,54 @@ export interface RecordStewardshipRequest {
...
+  readonly bagEmpty: boolean | null;
+  readonly bagFilled: boolean | null;
+  readonly tape: boolean | null;
+  readonly suckerGrowth: boolean | null;
+  readonly siteType: string | null;
+  readonly sidewalkWidth: string | null;
+  readonly siteWidth: string | null;
+  readonly siteLength: string | null;
+  readonly material: string | null;
+  readonly raisedBed: string | null;
```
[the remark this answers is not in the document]
SofieDunt: Updated documentation: https://github.com/Code-4-Community/c4c-dev-docs/pull/88

lnardini (approved):
Looks great! New routes seem like they fit in with the existing ones nicely :100:

chromium-52 (approved):
Looks great to me as well!! Thank you for working on this :)

vrushalitarte (approved):

PR #119, opened 2021-11-15
created upload sites form which takes in a csv and uploads those sites

willmt80:
DO NOT MERGE THIS OR FIX IT.

please try to make csv parsing work on the backend because this is awful. leaving the draft up for anyone who is curious. and specifically for the file upload and potentially file reading in case it's easier to pass raw csv as a string to the backend, but don't even try to convert to object or json here, that should absolutely be done with vertx.

## Why

[ClickUp Ticket](https://app.clickup.com/t/1qthfa3)

This PR allows admins to create new sites in bulk by uploading a single csv

## This PR

I am using the antd upload component to allow users to upload from their file explorer. This specifies that the file type must be a csv. For the parsing function, first the fileReader converts the csv from a file to a string, then the fast-csv parse converts that string to a javascript object.

The reason this was happening on the frontend was difficulty with file parsing on the backend. I may change this again to simply convert the file to a csv on the frontend then change the api call to accept a csv string as opposed to a file. I don't know.

## Screenshots

 

admin

## Verification Steps

 

I still need to write unit tests for the helper functions, hence the draft PR


chromium-52:
Closing in favor of #172

PR #120, opened 2021-12-30
Reports Page

SofieDunt:
## Why

[ClickUp Ticket](https://app.clickup.com/t/23k04f4)

Creates the report page for admins to view adoption and stewardship data. Adds a function for easier error handling ([new ticket to update existing code](https://app.clickup.com/t/23k0qcf)).

## This PR

- Connect the report routes to the api
- Create table components to display report entry info - paginated but hidden when only one page
- Create reports page, accessible only by admins, has featured stats and reports separated by tabs

## Screenshots

 

Data Loading
image

Data Loaded - Adoption Report Tab
image

Data Loaded - Stewardship Report Tab
image

## Verification Steps

 

Created tests where possible + ensured they passed, interacted with new page


vrushalitarte (approved):

src/components/tables/adoptionReportTable/index.tsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import { AdoptionReportEntry } from '../../../containers/reports/ducks/types';
+import { Table } from 'antd';
+import { dateSorter, DESCEND_ORDER, renderSiteIdAsLink } from '../utils';
+
+const AdoptionReportTableColumns = [
```
rymaju, line 6: 🏕️ We may want to bring this out into its own file for readability

src/components/tables/stewardshipReportTable/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { StewardshipReportTableEntry } from '../../../containers/reports/ducks/types';
+import { Table } from 'antd';
+import {
+  dateSorter,
+  DESCEND_ORDER,
+  renderActivities,
+  renderSiteIdAsLink,
+} from '../utils';
+
+const stewardshipReportTableColumns = [
```
rymaju, line 11: ⛺ Same with this

src/components/tables/test/utils.test.ts
```diff
@@ -0,0 +1,40 @@
+import { dateSorter, renderActivities, renderSiteIdAsLink } from '../utils';
+import { render, screen } from '@testing-library/react';
+
+test('dateSorter', () => {
```
rymaju, line 4: 🤔 This is the sort of function that a property test would be perfect for. If you're curious check out the fast-check package. Or take CS2800

src/components/tables/test/utils.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  expect(linkElement).toBeInTheDocument();
+  expect(linkElement).toHaveAttribute('href', '/tree/101');
+});
+
+test('renderActivities', () => {
+  render(renderActivities(['watered']));
+  let wateredTags = screen.getAllByText('watered');
+  expect(wateredTags.length).toEqual(1);
+
+  render(renderActivities(['watered', 'mulched']));
```
rymaju, line 28: ❓ Is it good practice/ok to call `render` multiple times in one test like this?

src/components/tables/utils.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const { Link } = Typography;
+
+export const DESCEND_ORDER: 'descend' | 'ascend' | null | undefined = 'descend';
+
+export function dateSorter(a: Date, b: Date): number {
+  return new Date(a).valueOf() - new Date(b).valueOf();
+}
+
+export function renderSiteIdAsLink(siteId: number): JSX.Element {
```
rymaju, line 13: ⛏️ These should be React components. The reason why you shouldn't return plain old JSX like this is because it messes with how React deals with updating the virtual DOM. I recommend this instead:
```ts
const SiteLink: React.FC<...> = ({siteId}) => {
```

src/containers/reports/ducks/selectors.ts
```diff
@@ -0,0 +1,47 @@
...
+  StewardshipReportTableEntry,
+} from './types';
+
+export const getCountAdoptedInPastWeek = (
+  adoptionReport: AdoptionReport | undefined,
+): number | undefined => {
+  return adoptionReport?.adoptionReport.filter(
+    (entry) =>
+      new Date(entry.dateAdopted) >
+      new Date(Date.now() - 7 * 24 * 60 * 60 * 1000),
```
rymaju, line 13: Can we have `new Date(Date.now() - 7 * 24 * 60 * 60 * 1000)` as a named constant for readability? Also do we need to use `Date.now()`, and if so, why no `new` keyword? In general we should construct all built-in classes like `Date`.

src/containers/reports/ducks/selectors.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    if (entry.mulched) {
+      activitiesPerformed.push('mulched');
+    }
+    if (entry.cleaned) {
+      activitiesPerformed.push('cleaned');
+    }
+    if (entry.weeded) {
+      activitiesPerformed.push('weeded');
+    }
```
rymaju, line 34: ⛏️ Seems like this should be pulled out into a helper. Its also a bit repetitive, we can instead for loop through the keys, check for their existence, and mutate. Or even better, filter a list of activities by those contained in `entry` like so:
```ts
const activitiesPerformed = ['watered', 'mulched', 'cleaned', 'weeded'].filter(activity => activity in entry)
```
Although using `in` to check for if a key exists in an object can be troublesome since it chases the prototype chain. You can also do:
```ts
const activitiesPerformed = ['watered', 'mulched', 'cleaned', 'weeded'].filter(activity => entry.hasOwnProperty(activity))
```

And and as a bonus we removed that nasty mutation 😊

src/utils/stringFormat.tsx
```diff
@@ -110,3 +110,11 @@ export function combineScientificName(entries: Entry[]): Entry[] {
 
   return newEntries;
 }
+
+/**
+ * Returns the error message of the given error.
+ * @param err the error
+ */
+export function getErrorMessage(err: any): string {
```
rymaju, line 118:
```suggestion
export function getErrorMessage(err: unknown): string {
```

rymaju (commented):
I have a few suggestions but nothing non-blocking. Overall looks good.

src/components/tables/test/utils.test.ts
SofieDunt: Agree, but I'd rather add this + improve other existing tests in a separate ticket. Left a todo.

src/components/tables/test/utils.test.ts
SofieDunt: The reason (that I could find) why it might be bad to call `render` multiple times is that it creates a new element each time, so if you're expecting a simple re-render you'd end up with multiple elements. In this case, that's what we want so it's fine. Still, I split this into two tests (to be more precise on what was being tested) which happens to remove the need for multiple renders.

src/components/tables/utils.tsx
SofieDunt: This was originally to avoid the `react/display-name` ESLint error when setting up the table columns. Created the React components and instead avoided the error by using JSX style API for the tables instead.

src/containers/reports/ducks/selectors.ts
SofieDunt: Created a named constant. `Date.now()` returns the number of ms since Jan 1, 1970 while `new Date()` returns the equivalent but as a Date object. We could use a date object in the calculation instead, but then we'd have to do `new Date().getDate()` to get the ms anyway, so I think using `Date.now()` is cleaner.

src/containers/reports/ducks/selectors.ts
SofieDunt: Was stumped for a while but figured out how to use filter and avoid the `TS7053` error (these checks are for whether the value of the key is true rather than if the key exists, so had to use `entry[activity]` instead of `entry.hasOwnProperty(activity)`. This gives you a type error unless you explicitly declare that the list is made up of valid keys).

src/utils/stringFormat.tsx
SofieDunt: Can't access `err.response.data` if `err` is `unknown`. Could make an interface for errors instead? (would prefer to do this in a separate ticket)

src/components/tables/adoptionReportTable/index.tsx
SofieDunt: Used [JSX style API](https://ant.design/components/table/#components-table-demo-jsx) instead.

rymaju (approved):
LGTM

PR #122, opened 2022-01-27
Reflect updated back end naming convention

SofieDunt:
## Why

[ClickUp Ticket](https://app.clickup.com/t/28ry9d3)

SFTT noticed the neighborhoods of the map all had the same opacity.

## This PR

Found the bug was caused because the name passed to `getProperty` did not exist, making the method return undefined and thus setting the opacity to the default for all neighborhoods. This is a result of the updated back end naming conventions. Changed names passed to `getProperty` to camelCase to reflect the change from snake_case in the back end.

## Screenshots

 

Properly styled neighborhoods:
image

## Verification Steps

 
Visually confirmed the neighborhood opacities are correct.

### DIFF
--- a/src/components/mapComponents/logic/event.ts
+++ b/src/components/mapComponents/logic/event.ts
@@ -77,7 +77,7 @@ export function addShowReservationModal(): void {
   blocksLayer.addListener('click', (event) => {
     // get status of block based on color
     const status: ReservationModalType = ((): ReservationModalType => {
-      switch (event.feature.getProperty('block_id') % 10) {
+      switch (event.feature.getProperty('blockId') % 10) {
         case 1:
           return ReservationModalType.TAKEN;
         case 0:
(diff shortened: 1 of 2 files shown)


florisdobber (approved):
Nice catch 🎣

src/components/mapComponents/logic/event.ts
```diff
@@ -77,7 +77,7 @@ export function addShowReservationModal(): void {
   blocksLayer.addListener('click', (event) => {
     // get status of block based on color
     const status: ReservationModalType = ((): ReservationModalType => {
-      switch (event.feature.getProperty('block_id') % 10) {
+      switch (event.feature.getProperty('blockId') % 10) {
```
rymaju, line 80: ⛏️ Can we make this a named constant across the project?

rymaju (approved):

PR #123, opened 2022-01-31
Refresh Updates Redux Store

SofieDunt:
## Why

[ClickUp Ticket](https://app.clickup.com/t/1x52pz7)

Ensures the access token is refreshed or the user is logged out when access token expires.

## This PR

Previously, our way of refreshing never actually updated the store with the fresh access token–instead, axios would always catch the expired access token error, get a new access token, and apply that token to the next call (re-making the original request). This ticket creates a refresh thunk to update the Redux store via dispatch. The axios response interceptor uses the refresh thunk and logout thunk instead of calling authClient directly to ensure the store is updated appropriately.

## Verification Steps

 

Wrote tests for the new refresh thunk and ran the front end/back end to check that refresh worked after the access token expired.

### DIFF
--- a/src/auth/ducks/thunks.ts
+++ b/src/auth/ducks/thunks.ts
@@ -1,5 +1,6 @@
 import {
   LoginRequest,
+  RefreshTokenResponse,
   SignupRequest,
   TokenPayload,
   UserAuthenticationThunkAction,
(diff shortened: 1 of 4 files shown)


src/auth/ducks/thunks.ts
```diff
@@ -29,6 +30,27 @@ export const login = (
...
+): UserAuthenticationThunkAction<void> => {
+  return (dispatch, getState, { authClient }): Promise<void> => {
+    dispatch(authenticateUser.loading());
+    return authClient
+      .refresh(refreshToken)
+      .then((refreshTokenResponse: RefreshTokenResponse) => {
+        dispatch(
+          authenticateUser.loaded({
+            accessToken: refreshTokenResponse.freshAccessToken,
+            refreshToken,
```
rymaju, line 44: Nice! Im assuming this sets the store correctly

src/auth/test/thunks.test.ts
```diff
@@ -84,6 +84,71 @@ describe('User Authentication Thunks', () => {
     });
   });
 
+  describe('refresh', () => {
```
rymaju, line 87: Its beautiful 😭

rymaju (approved):
LGTM Im sad this wasnt done right the first time

chromium-52 (approved):
Looks great!!

PR #124, opened 2022-02-01
Copy email in reports to clipboard

SofieDunt:
## Why

[ClickUp Ticket](https://app.clickup.com/t/28t1r2h)

Allows admins to easily copy the emails listed in the reports.

## This PR

Created a new component, ClickableText, that (by default) copies the displayed text to the clipboard.
- Uses the [ClipboardAPI](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API), which is not supported on Internet Explorer - users will be notified if their browser doesn't support copying
- Takes `onCopy` as a prop for extension + testing purposes

## Screenshots

 

image

## Verification Steps

 
Wrote tests to check the logic and error handling.

### DIFF
--- a/src/components/copyText/index.tsx
+++ b/src/components/copyText/index.tsx
@@ -7,6 +7,10 @@ interface CopyTextProps {
   readonly onCopy?: (text: string) => Promise<void>;
 }
 
+/**
+ * Writes to the system clipboard using the [Clipboard Api](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API).
+ * @param text the string to copy to the clipboard
+ */
 export const copyToClipboard = (text: string): Promise<void> => {
   return navigator.clipboard.writeText(text);
 };
--- a/src/components/copyText/index.tsx
+++ b/src/components/copyText/index.tsx
@@ -1,0 +1,44 @@
+import React, { useState } from 'react';
+import { Button, Tooltip } from 'antd';
+import styled from 'styled-components';
+
+interface CopyTextProps {
+  readonly text: string;
+  readonly onCopy?: (text: string) => Promise<void>;
+}
+
+export const copyToClipboard = (text: string): Promise<void> => {
+  return navigator.clipboard.writeText(text);
+};
+
+const ClickableText = styled(Button)`
+  border: none;
+  padding: 0;
+  background: inherit;
+  box-shadow: none;
+
+  &:hover {
+    background: inherit;
+  }
+`;
+
+const CopyText: React.FC<CopyTextProps> = ({
+  text,
+  onCopy = copyToClipboard,
+}) => {
+  const [copiedFeedback, setCopiedFeedback] = useState('Copied to clipboard!');
+
+  const onClick = () => {
+    onCopy(text).catch(() => {
+      setCopiedFeedback('Browser does not support copying to clipboard.');
+    });
+  };
+
+  return (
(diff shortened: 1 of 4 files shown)


src/components/copyText/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { Button, Tooltip } from 'antd';
+import styled from 'styled-components';
+
+interface CopyTextProps {
+  readonly text: string;
+  readonly onCopy?: (text: string) => Promise<void>;
+}
+
+export const copyToClipboard = (text: string): Promise<void> => {
+  return navigator.clipboard.writeText(text);
```
rymaju, line 11: ‼️ Where does navigator come from? It looks like its a window scoped global variable... you know how I feel about those 🤢

src/components/copyText/test/copyText.test.tsx
```diff
@@ -0,0 +1,50 @@
+import { act, fireEvent, render, screen } from '@testing-library/react';
+import CopyText, { copyToClipboard } from '../index';
+import React from 'react';
+
+Object.assign(navigator, {
```
rymaju, line 5: In this case, you made onCopy a function you can pass in, why not just pass in a jest.fn instead of mutating a global var? It makes the tests a lot more readable

src/components/copyText/index.tsx
rymaju: ⛏Would also prefer if you left a comment in the code explaining this a bit more

src/components/copyText/test/copyText.test.tsx
rymaju: Ah I see nevermind, see the comment below

src/components/copyText/test/copyText.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+  clipboard: {
+    writeText: jest.fn(),
+  },
+});
+
+describe('copyText tests', () => {
+  describe('copyToClipboard test', () => {
+    it('writes to clipboard', async () => {
```
rymaju, line 13: In the future, you dont need to test something like this, but im glad you did anyway!

src/components/copyText/test/copyText.test.tsx
rymaju: Maybe a line describing what this is doing would help here

src/components/tables/adoptionReportTable/index.tsx
```diff
@@ -32,6 +33,7 @@ const AdoptionReportTable: React.FC<AdoptionReportTableProps> = ({
         title={"Adopter's Email"}
         dataIndex={'email'}
         key={'email'}
+        render={(email: string) => <CopyText text={email} />}
```
rymaju, line 36: Weird that theres no AntD copy text component, but if it works!

Is there a visual indication that you copied something? ❓

rymaju (approved):
Have a few questions but nothing blocking, as always I love the effort put into testing

src/components/copyText/index.tsx
SofieDunt: It might be... there's really no way around it though. We could install [react-copy-to-clipboard](https://www.npmjs.com/package/react-copy-to-clipboard), but that uses `window.clipboardData.setData(format, text)` under the hood anyway.

zhangcry (approved):

PR #125, opened 2022-02-13
added reports link to homepage for admins

jhylisa666:
## Why
added reports link and display box to the homepage if the user is an admin by adding conditionals(ternary operators)

If the given user is an admin, then append REPORT to the list of display, otherwise, show the default display.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 # IDE files
 .idea
 .vscode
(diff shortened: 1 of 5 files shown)


Tvcz (approved):
👍

SofieDunt (approved):
Looks great! 🔥

PR #126, opened 2022-02-16
Improve after-login redirect

SofieDunt:
## Why

[ClickUp Ticket](https://app.clickup.com/t/2b93c6u)

After a user logs in, they're redirected from the Login page to their original destination (or the Home page if they had none). The way this originally was done meant trying to render the Login page while updating the state and rendering the new page (basically, a mess). 

This ticket improves how we handle redirect and login to get rid of this error.

## This PR

- Instead of using `useHistory` to push the user to a new location, the Login page simply returns a `Redirect` (and, more importantly, doesn't return the login page at the same time!)
- We have to `getUserData` to show the user's name on the home page and in the nav bar. Originally, we did this in the login page after they logged in. However, it makes a lot more sense to get that user data before the user is redirected to the new page (ex. if the data doesn't load in time, the name will just be missing). This ticket adds a call to get the user data in the `login` thunk so that we always get the user data at the same time as log in (and don't rely on a page/component to know it has to call `getUserData` too).
  - note: we use async in the `login` thunk to ensure that we finish getting the user data before letting the rest of the app know that we've finished logging in

## Verification Steps

 

You can run the front end and try logging in to see that redirect works, the user data is fetched successfully, and the error above doesn't show up.

I also modified some old tests to check that `getUserData` is successfully called in the login thunk and created some new integration tests to ensure that the Login page renders correctly based on the `authenticationState`.

### DIFF
--- a/src/containers/login/index.tsx
+++ b/src/containers/login/index.tsx
@@ -1,10 +1,10 @@
 import React from 'react';
 import { useLocation } from 'react-router';
-import { Link, useHistory } from 'react-router-dom';
+import { Link, Redirect } from 'react-router-dom';
 import { connect, useDispatch, useSelector } from 'react-redux';
 import { Helmet } from 'react-helmet';
 import { C4CState } from '../../store';
-import { login, getUserData } from '../../auth/ducks/thunks';
+import { login } from '../../auth/ducks/thunks';
 import {
   LoginRequest,
   UserAuthenticationReducerState,
@@ -97,7 +97,6 @@ interface LoginProps {
 const Login: React.FC<LoginProps> = ({ tokens }) => {
   const { windowType } = useWindowDimensions();
   const dispatch = useDispatch();
-  const history = useHistory();
   const location = useLocation<RedirectStateProps>();
   const loggedIn = useSelector((state: C4CState) =>
     isLoggedIn(state.authenticationState.tokens),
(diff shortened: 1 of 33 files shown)


src/containers/login/index.tsx
```diff
@@ -139,101 +131,105 @@ const Login: React.FC<LoginProps> = ({ tokens }) => {
...
-                        onFinish={onFinish}
-                        windowType={windowType}
-                      />
-                      {ForgotPasswordFooter}
-                    </InputContainer>
-
-                    <Col span={2} />
-
-                    <Col span={12}>
+  if (loggedIn) {
```
rymaju, line 134: ⛏️ I like ternaries more but this is good too!

src/containers/login/test/login.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+} from '../../../utils/asyncRequest';
+import { Store } from 'redux';
+import { configureStore } from '@reduxjs/toolkit';
+import { BrowserRouter, Route } from 'react-router-dom';
+import { Routes } from '../../../App';
+import { mockTokenResponse, mockUserDataResponse } from '../../../App.test';
+
+// window.matchMedia is invoked when rendering the Login page, but is not implemented in JSDOM, so we mock it here
+// see more: https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom
+Object.defineProperty(window, 'matchMedia', {
```
rymaju, line 19: 💭 It sucks we need to do this, but this is why global vars suck 😔

rymaju (approved):
LGTM

Tvcz (approved):
After reading this over a few times, I think I can say with confidence that this is indeed code.

PR #127, opened 2022-02-19
Open tree page in a new tab

chromium-52:
## Why

[ClickUp Ticket](https://app.clickup.com/t/1x52qae)

Open the tree page in a new tab after clicking "More Info" on the map popup.

## This PR

Pass in "_blank" as `target`. The change that led to this bug was through #113 (just a little confused as to why this change was initially required - would this ticket break how the popup and return button function on mobile?).

## Screenshots

 
Clicking the "More Info" button on the tree popup opens tree page in a new tab (at least on desktop).
image

Clicking on the "Return to Tree Map" button redirects to the main page with the map with the right location and street-level zoom.
image

## Verification Steps

Tested that the tree page opens up in a new page and that clicking the "Return to Tree Map" button redirects to the map page with the correct location. Not tested on mobile.

### DIFF
--- a/src/components/mapComponents/maps/mapWithPopup/index.tsx
+++ b/src/components/mapComponents/maps/mapWithPopup/index.tsx
@@ -186,11 +186,7 @@ const MapWithPopup: React.FC<MapWithPopupProps> = ({
         />
       </div>
       <MapDiv id="map" ref={mapRef} />
-      <TreePopup
-        treeInfo={activeTreeInfo}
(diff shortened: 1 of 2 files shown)


SofieDunt (approved):
👍 Makes sense to me

rymaju (approved):
LGTM

chromium-52:
Just tested it on mobile and it looks good to me.

PR #128, opened 2022-03-20
Improve forgot password page

chromium-52:
## Why

[ClickUp Ticket](https://app.clickup.com/t/1gewz96)

## This PR

Improve the forgot password page so that it displays appropriate messages instead of alert popups and console logs when the password reset request is successful or not. Upon a successful request, update the page to prevent the user from submitting multiple requests.

## Screenshots

 
Initial screen:
image

User submits non-existent email:
image

User submits correct email:
image

## Verification Steps

 
Verify that the messages shown are as expected when the user submits incorrect and correct email addresses.

### DIFF
--- a/src/containers/forgotPassword/index.tsx
+++ b/src/containers/forgotPassword/index.tsx
@@ -12,19 +12,13 @@ import { enterEmailRules } from '../../utils/formRules';
 const ForgotPassword: React.FC = () => {
   const { windowType } = useWindowDimensions();
 
-  const [submittedEmail, setSubmittedEmail] = useState(false);
   const [email, setEmail] = useState('');
 
   const onFinish = (values: ForgotPasswordRequest) => {
-    authClient
-      .forgotPassword(values)
-      .catch(() => {
-        return;
-      })
-      .finally(() => {
-        setSubmittedEmail(true);
-        setEmail(values.email);
-      });
+    setEmail(values.email);
+    authClient.forgotPassword(values).catch(() => {
+      return;
+    });
   };
 
   return (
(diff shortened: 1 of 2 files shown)


src/containers/forgotPassword/index.tsx
```diff
@@ -33,19 +40,39 @@ const ForgotPassword: React.FC = () => {
...
+        {!submittedEmail || userNotFound ? (
+          <>
+            <Typography.Title level={4}>
+              Please enter your email and we will send you the password reset
+              instructions to the email address for this account.
+            </Typography.Title>
+            <Form form={form} name="forgotPassword" onFinish={onFinish}>
+              {userNotFound && (
+                <Typography.Title level={4}>
+                  We couldn't find an account associated with the given email.
```
chromium-52, line 55: I don't feel too good showing this message to the users because it technically allows attackers to potentially submit a bunch of different emails to see which ones are in use but I'm also not sure how common this would even be and/or how valuable information gained from this message would be to attackers.

rymaju: I agree, this is a big yikes from a security point of view.

SofieDunt: Good point, you can switch this to a message that just says "If an account is registered under [given email], we have sent you a link to reset your password."

🔧 Could you also change the `LOGIN_ERROR` message under assets/content.tsx to 'The email or password you entered was incorrect'? Not in the scope of this ticket, but it should be a quick fix and is relevant to this idea of hiding what emails are registered.

src/containers/forgotPassword/index.tsx
```diff
@@ -12,16 +12,21 @@ import { enterEmailRules } from '../../utils/formRules';
 const ForgotPassword: React.FC = () => {
   const { windowType } = useWindowDimensions();
 
+  const [submittedEmail, setSubmittedEmail] = useState(false);
+  const [email, setEmail] = useState('');
```
SofieDunt, line 16: ⛏️ Could maybe just have `email` and check if it's null or empty string to check whether or not they've submitted the email.

SofieDunt (approved):
LGTM 👍

Tvcz (approved):
lgtm

PR #129, opened 2022-03-29
Scroll Bars and Reflexive Styling

SofieDunt:
## Why

[ClickUp Ticket](https://app.clickup.com/t/2my7fvm)

Improves the user experience by removing unnecessary scroll bars and improving reflexive styling. Also fixed the line not extending across the entire tree popup.

## This PR

- AntD's components have been too rigid for reflexive styling. I replaced Rows, Cols, and Lists with flex divs. 
- From there, it's mostly css changes to height/width/margins/etc. to make everything look good. 
- Used css media queries at times to simplify styling instead of hooking into the state as we've been doing - for the future, we'll reserve use of the hook to component layouts rather than styled component styling.
- Use message instead of Alerts to inform users of login/signup errors - this makes styling much easier and avoids the error we had where it's impossible to distinguish between a login error and a signup error.

## Screenshots

 
Tree Page Info (desktop)
image

Tree Page Info (desktop, zoomed out)
image

Login Page (desktop)
image

Login Page (tablet)
image

Login Error (narrow desktop)
image

Signup Error (narrow desktop)
image

## Verification Steps

 
Made sure everything looked reasonable on various screen sizes in Google Chrome, Safari, and Firefox (tested on a Mac).

### DIFF
--- a/src/auth/ducks/thunks.ts
+++ b/src/auth/ducks/thunks.ts
@@ -14,6 +14,7 @@ import Client from '../../api/protectedApiClient';
 
 export const login = (
   loginRequest: LoginRequest,
+  onError: () => void,
 ): UserAuthenticationThunkAction<void> => {
   return (
     dispatch,
@@ -37,6 +38,7 @@ export const login = (
       })
       .catch((error) => {
         dispatch(authenticateUser.failed(error.response.data));
+        onError();
       });
   };
 };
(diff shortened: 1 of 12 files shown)


Tvcz (approved):
LGTM, though the unused values will have to be removed before the checks pass.

I tested the login, signup, and tree information pages on Google Chrome (Windows), and it looks good with all screen sizes.

PR #131, opened 2022-04-03
Display only streets and parks in map

chromium-52:
## Why

[ClickUp Ticket](https://app.clickup.com/t/27pkmw3)

Remove names of buildings and other text so that the map mostly just displays street and park names.

## This PR

Change the options for the map so that it doesn't display names of POIs (points of interest) except for parks.

## Screenshots

 
image

## Verification Steps

 
Visually tested that the map mostly only displays street and park names on both desktop and mobile in Chrome and Firefox.

### DIFF
--- a/src/components/mapComponents/constants.tsx
+++ b/src/components/mapComponents/constants.tsx
@@ -64,6 +64,8 @@ export const SITE_OPTIONS: { label: ReactNode; value: string }[] = [
   },
 ];
 
+// Relevant documentation: https://developers.google.com/maps/documentation/javascript/style-reference
+// Removes the map's labels for all POIs (points of interest) such as businesses and schools except parks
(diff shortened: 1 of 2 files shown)


Tvcz (approved):
lgtm

src/components/mapComponents/constants.tsx
```diff
@@ -63,3 +63,22 @@ export const SITE_OPTIONS: { label: ReactNode; value: string }[] = [
     value: 'Open',
   },
 ];
+
+export const LIGHT_MAP_STYLES: google.maps.MapTypeStyle[] = [
```
SofieDunt, line 67: ⛏️ Can you leave a comment here linking the google maps documentation we used and explaining what we're hiding here?

SofieDunt (approved):
💯

PR #138, opened 2022-04-14
Auth Error Interceptor

SofieDunt:
## Why

[ClickUp Ticket](https://app.clickup.com/t/29a81x7)

Doesn't show auth error after successful refresh and properly logs out user if refresh token is expired. Adds an error type to make handling errors easier.

## This PR

Essentially made some fixes to responseErrorInterceptor - will leave comments.

## Verification Steps

 
- Wrote tests for the error response interceptor and ensured they passed
- Manually checked that refresh/force logout worked (you can make this easier to test by changing `expiration_ms_refresh` and `expiration_ms_access` in the backend)

### DIFF
--- a/src/auth/axios.ts
+++ b/src/auth/axios.ts
@@ -1,11 +1,12 @@
 import axios, { AxiosError, AxiosInstance, AxiosResponse } from 'axios';
-import store from '../store';
+import store, { LOCALSTORAGE_STATE_KEY } from '../store';
 import { asyncRequestIsComplete } from '../utils/asyncRequest';
 import { UserAuthenticationReducerState } from './ducks/types';
 import { isTokenValid } from './ducks/selectors';
 import { logout, refresh } from './ducks/thunks';
 import authClient from './authClient';
 import protectedApiClient from '../api/protectedApiClient';
+import history from '../history';
 
 const AppAxiosInstance: AxiosInstance = axios.create({
   baseURL: process.env.REACT_APP_API_DOMAIN,
@@ -58,7 +59,9 @@ export const responseErrorInterceptor = async (
     error?.response?.data === INVALID_ACCESS_TOKEN &&
     !isTokenValid(tokens.result.refreshToken)
   ) {
-    logout()(store.dispatch, store.getState, userAuthenticationExtraArgs);
+    await logout()(store.dispatch, store.getState, userAuthenticationExtraArgs);
+    localStorage.removeItem(LOCALSTORAGE_STATE_KEY);
+    history.go(0);
   }
   return Promise.reject(error);
 };
--- a/src/auth/axios.ts
+++ b/src/auth/axios.ts
@@ -1,4 +1,4 @@
-import axios, { AxiosError, AxiosInstance } from 'axios';
+import axios, { AxiosError, AxiosInstance, AxiosResponse } from 'axios';
 import store from '../store';
 import { asyncRequestIsComplete } from '../utils/asyncRequest';
 import { UserAuthenticationReducerState } from './ducks/types';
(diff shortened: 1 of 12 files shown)


src/auth/axios.ts
```diff
@@ -38,7 +46,7 @@ const responseErrorInterceptor = (error: AxiosError) => {
     isTokenValid(tokens.result.refreshToken) &&
     !(error.config as any)?._retry
   ) {
-    refresh(tokens.result.refreshToken)(
+    await refresh(tokens.result.refreshToken)(
```
SofieDunt, line 49: 📝 If you don't wait, the original request will be returned and made again with the old access token, resulting in yet another unauthorized access error.

src/auth/axios.ts
```diff
@@ -51,7 +59,9 @@ const responseErrorInterceptor = (error: AxiosError) => {
     error?.response?.data === INVALID_ACCESS_TOKEN &&
     !isTokenValid(tokens.result.refreshToken)
   ) {
-    logout()(store.dispatch, store.getState, userAuthenticationExtraArgs);
+    await logout()(store.dispatch, store.getState, userAuthenticationExtraArgs);
+    localStorage.removeItem(LOCALSTORAGE_STATE_KEY);
```
SofieDunt, line 63: 📝 Honestly not entirely sure why we have to clear the storage again - I assume it has something to do with the context.

src/auth/axios.ts
```diff
@@ -51,7 +59,9 @@ const responseErrorInterceptor = (error: AxiosError) => {
...
+    history.go(0);
```
SofieDunt, line 64: 📝 Force the page to reload, which will make the user go to the login page as App will recognize they are no longer logged in

src/auth/test/axios.test.ts
```diff
@@ -0,0 +1,132 @@
...
+};
+
+// a mock partial state with an invalid refresh token
+const mockInvalidAuthState: Partial<C4CState> = {
+  authenticationState: {
+    tokens: AsyncRequestCompleted(mockTokenResponse(invalidExp)),
+    userData: AsyncRequestCompleted(mockUserDataResponse),
+  },
+};
+
```
SofieDunt, line 34: 📝 Fun fact - if you `jest.mock('../../store')` here and `npm run test`, tests in other modules will break. Presumably this is because all the modules depend on the mock, but running `jest.resetAllMocks` and `jest.unmock` before each test won't work...  My takeaway here is to mock as conservatively as possible.

src/containers/reports/index.tsx
```diff
@@ -31,18 +31,21 @@ const Reports: React.FC = () => {
...
+        ProtectedApiClient.getStewardshipReport()
+          .then((stewardshipRes: StewardshipReport) =>
+            setStewardshipReport(stewardshipRes),
+          )
+          .catch((err) =>
+            message.error(
+              `Could not get stewardship report: ${getErrorMessage(err)}`,
+            ),
+          );
+      })
```
SofieDunt, line 45: 📝 So you won't get two auth errors at once

src/auth/axios.ts
```diff
@@ -22,7 +23,14 @@ const userAuthenticationExtraArgs = {
 
 const INVALID_ACCESS_TOKEN = 'Given access token is expired or invalid';
 
-const responseErrorInterceptor = (error: AxiosError) => {
+// TODO Use throughout application
```
rymaju, line 26: 💭 Could you elaborate on this more? How will we go about adding this and changing old code?

src/auth/axios.ts
rymaju: If on a public computer then its possible to grab the token in a XSS and login to your account.

src/auth/test/axios.test.ts
rymaju: Yes! I personally hate the jest.mock(module) function.

We should ideally only use jest for its cool test functions and nothing else. Pass stateful dependencies as parameters!!

rymaju (approved):
LGTM

zhangcry (approved):

src/auth/axios.ts
SofieDunt: Good point, added elaboration in [this ticket](https://app.clickup.com/t/23k0qcf)

PR #140, opened 2022-04-23
Improve after-signup redirect

SofieDunt:
## Why

[ClickUp Ticket](https://app.clickup.com/t/27qb23u)

After a user logs in, they're redirected from the Signup page to the Home page. The way this originally was done meant trying to render the Signup page while updating the state and rendering the new page.

This ticket improves how we handle redirect and signup to get rid of this error (https://github.com/Code-4-Community/speak-for-the-trees-frontend/pull/126 part 2).

Navigating to signup/login through navbar also now redirects you back to your original page after auth for convenience (instead of home).

## This PR

- Instead of using useHistory to push the user to a new location, the Signup page simply returns a Redirect
- Removes unused props
- Uses location to push the original destination in navbar link buttons

## Verification Steps

 
You can run the front end and try signing up to see that redirect works and the error doesn't show up.

I also created tests to ensure that the Signup page renders correctly based on the authenticationState and that the signup/login buttons successfully push the destination in the state.


Tvcz (approved):
Tested with a few signups and logins with no error. Redirects to the expected page when logging in or using the "sign up here" link on the login screen. LGTM 👍

Tvcz (commented):
:thought_balloon: One thing I did notice when prompted to login or signup is that the redirect only works if you click the "sign up here" link on the login screen and not if you click the "sign up" button in the top-right. Though this could plausibly be behavior as intended, I thought it was worth noting.

zhangcry (approved):

PR #155, opened 2022-07-30
Fix 4-character Zip Codes

chromium-52:
## Why

[ClickUp Ticket](https://app.clickup.com/t/29xbk00)

Change `zipCodeRules` to allow sites to be created/edited with a 5-character numerical zip code (instead of 4). Will update the database so that all zip codes are 5 characters long when this PR gets merged.

## This PR

Change `zipCodeRules`  so that zip codes must be 5-character numeric values. Also, noticed that there's no input validation in the frontend on submit. The neighborhood dropdown also doesn't display the correct option. Will handle both in separate PRs.

## Screenshots

 
image
image
zip-code-leading-zero

## Verification Steps

 
Verified that the zip code field is green only when given a valid input (i.e. 5-digit number) by testing various cases (e.g. 'a', '1234', '1234a', '123456').
Verified that the zip code updates correctly when clicking on the submit button.

### DIFF
--- a/src/components/treeInfo/index.tsx
+++ b/src/components/treeInfo/index.tsx
@@ -53,7 +53,7 @@ const TreeInfo: React.FC<TreeProps> = ({
     // TODO change to siteData.city and remove check for zip after data is cleaned
     let baseLocation = `Boston`;
     if (siteData.zip) {
-      baseLocation += ` 0${siteData.zip}`;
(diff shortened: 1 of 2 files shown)


Tvcz (approved):
lgtm!

zhangcry (approved):

PR #157, opened 2022-08-08
Default form options to previous site entry and turn status input field into radio button

chromium-52:
## Why

Set the default value for the radio buttons in the add site entry form to be the value of the previous (i.e. latest) entry for that site.
Turn `status` into enum `SiteEntryStatus` from `string` and replace the status text field with a radio button.

## This PR

The default values for all of the radio buttons are `false` with the exception of `status`, which is `null`.
Create an enum of three values that match the three possible values for `status` (i.e. Alive, Dead, or Dead but standing). If the status radio button isn't selected, the status will be set to `null`.

## Screenshots

 
localhost_3000_site_3331239

## Verification Steps

 
Check that the default values for the radio buttons are correct both when `latestSiteEntry` is and isn't given (including in the `/admin` page).


Tvcz (approved):
LGTM

PR #147, opened 2022-05-01
Edit site page link

aschimmichanga:
## Why

[ClickUp Ticket](https://app.clickup.com/xxxxxxxx/v/b/li/xxxxxxxxx)

Allows admins to easily edit tree site pages without changing experience of normal users

## This PR

Needed to change the way that sites that do not have trees are represented throughout the codebase
Needed to implement a button that conditionally renders in the tree popup only if the user has admin privileges

## Screenshots

# Admin View:
image
image

# User View
image
image

# Page that shows up after clicking button
image

## Verification Steps
Logged in as normal user/admin respectively
Zoomed into tree map
Clicked on yellow and green icons to bring up site/tree popup
Clicked on "Edit Site Page" button to bring up the editable Site Page

PR #158, opened 2022-08-14
Increase MyTrees sidebar width for desktop devices

chromium-52:
## Why

Increase width of the MyTrees sidebar so that both the More Info and Edit Site Page buttons are shown in a single row (only an issue with admin users). Bug introduced in #147 (this is on me)

## This PR

Change the width of the sidebar from 20vw to 22wv

## Screenshots

 
Before:
image

After:
image

## Verification Steps

 
Checked that the buttons are rendered correctly in both desktop, tablet, and mobile devices


zhangcry (approved):

priime0 (approved):
lgtm

PR #159, opened 2022-10-18
add share buttons to my trees menu and tree info pages

huang0h:
## Why

[ClickUp Ticket](https://app.clickup.com/t/1dk497r)

Add buttons to my trees/tree info pages so users can easily share tree pages through social media and email

## This PR

- Add media share icon images to assets (twitter logo, email icon, etc.)
- Create new component shareButton to display/hide the menu containing media share buttons
- Create new component shareMenu to house media share buttons, message, and sharing logic
- Add shareButton to treeCard and treeInfo components
- Create a new styled component UnadoptButton to recolor the unadopt button on tree info pages to red/white - helps convey negative action/signal to be careful

## Screenshots

On my trees (before vs. after):
image
image

On a tree's info page (before vs. after):
image
image

## Verification Steps

- Checked that default messages are properly displayed based on if the user has adopted the tree or if someone else has already adopted it
- Checked that the media share buttons properly generate posts on click with only the content in the input box (or just the link, in the case of the facebook button)
- Checked that the posts generated by the media share buttons are valid and can actually be posted/sent
- Checked that the share button (the one with the arrow coming out of the square) and content in the share menu display correctly on desktop, tablets, and mobile devices


src/components/shareMenu/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+import EmailIcon from '../../assets/images/email-icon.png';
+import FBIcon from '../../assets/images/facebook-icon.png';
+import CopyIcon from '../../assets/images/copy-icon.png';
+
+const ShareMenuContainer = styled.div`
+  max-width: 550px;
+  margin: 20px 0px;
+`;
+
+const StyledButton = styled(Button)`
```
priime0, line 14: nit: `StyledButton` doesn't really convey the intent of how and where it should be used. Maybe we should rename it?

src/components/shareButton/index.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { useState } from 'react';
+import styled from 'styled-components';
+import { Button } from 'antd';
+import { SizeType } from 'antd/es/config-provider/SizeContext';
+import ShareMenu from '../shareMenu';
+import ShareWhite from '../../assets/images/sharewhite-small.png';
```
priime0, line 6: nit: I personally find the other share icons more intuitive (arrow to top right from window, 3 node tree) and didn't know what this button meant at first. It might just be me though.

src/components/treeInfo/index.tsx
```diff
@@ -63,6 +76,25 @@ const TreeInfo: React.FC<TreeProps> = ({
...
+    <ShareButton
+      size={mobile ? 'middle' : 'large'}
+      // if user owns the tree, show first line
+      // otherwise if someone else has adopted, show second line
+      // otherwise nobody has adopted the tree - show final line
+      defaultText={
+        userOwnsTree
+          ? 'Check out this tree I adopted!'
+          : adopted
+          ? 'Check out this tree near you!'
```
priime0, line 89: nit: Maybe this should be resolved in the ticket, but I think "Adopt this tree near you!" might be better.

priime0 (approved):
lgtm :+1:

chromium-52 (commented):
:100: lgtm! Out of curiosity, is it possible to add insta to the list of socials or would that be too tricky?

src/components/shareButton/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const StyledShareButton = styled(Button)`
+  && {
```
chromium-52, line 9: :question: Are the `&&`s necessary? This seems to be an equivalent of `!important` and we generally want to use it sparingly

src/components/treeCard/index.tsx
```diff
@@ -55,6 +56,11 @@ const TreeCard: React.FC<TreeCardProps> = ({ site }) => {
           >
             More Info
           </GreenLinkButton>
+          <ShareButton
+            size="middle"
+            defaultText="Check out this tree I adopted!"
+            link={`map.treeboston.org/tree/${site.id ? site.id : ''}`}
```
chromium-52, line 62: :memo: Just a random JS trivia that `site.id ? site.id : ''` can be simplified to `site.id || ''` (or `site.id ?? ''` depending on the situation) but this is fine too

chromium-52 (commented):
:100: lgtm! Out of curiosity, is it possible to add insta to the list of socials or would that be too tricky?

src/components/treeInfo/index.tsx
huang0h: not sure which case you're talking about - if you mean the second line ("Check out this tree near you!"), that's for trees that someone else has already adopted

src/components/shareMenu/index.tsx
huang0h: yeah that's fair - renaming to MediaShareButton

src/components/shareButton/index.tsx
huang0h: This one is on the more abstract side, but it's also not uncommon - I believe Instagram and Twitter use this icon, just to name a few, so there's a good chance people have seen it before. To me, the arrow implies a reply, although that doesn't make sense in this context, and I haven't seen the 3-node tree around that much lately - I think really any of these could work, though

src/components/treeInfo/index.tsx
priime0: Ah that makes sense.

src/components/shareButton/index.tsx
priime0: Yup, that works then.

huang0h:
> 💯 lgtm! Out of curiosity, is it possible to add insta to the list of socials or would that be too tricky?

I don't think so - looks like there's the [graph api](https://developers.facebook.com/docs/instagram-api) businesses can use to manage their own accounts, and the [basic display api](https://developers.facebook.com/docs/instagram-basic-display-api) that can only retrieve data. I did just find a way to share links through facebook messenger (https://developers.facebook.com/docs/sharing/reference/send-dialog), but we'd have to set up the [facebook SDK](https://developers.facebook.com/docs/javascript) to do so

src/components/shareButton/index.tsx
huang0h: Took another look, they're not needed - will keep that in mind!

huang0h:
> 💯 lgtm! Out of curiosity, is it possible to add insta to the list of socials or would that be too tricky?

I don't think so - looks like there's the [graph api](https://developers.facebook.com/docs/instagram-api) businesses can use to manage their own accounts, and the [basic display api](https://developers.facebook.com/docs/instagram-basic-display-api) that can only retrieve data. I did just find a way to share links through facebook messenger (https://developers.facebook.com/docs/sharing/reference/send-dialog), but we'd have to set up the [facebook SDK](https://developers.facebook.com/docs/javascript) to do so

src/components/shareButton/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+  padding-left: 10px;
+  padding-right: 10px;
+`;
+
+const StyledImg = styled.img`
+  max-height: 25px;
+  margin-bottom: 4px;
+`;
+
+// link prop is primarily for facebook as the post can only be prefilled with a link, no body text
```
rymaju, line 19: Not everything should be represented as a single object. We have an enumeration here: one repr per social media. What does FI/FII tell us about how to design such data and behavior?

src/components/shareMenu/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+interface ShareMenuProps {
+  defaultText: string;
+  link: string;
+}
+
+const ShareMenu: React.FC<ShareMenuProps> = ({ defaultText, link }) => {
+  const [formText, setFormText] = useState(`${defaultText} ${link}`);
+
+  // goal is to open email window in new tab to stay consistent with more info and other share buttons
```
rymaju, line 37: Code is not a place for streams of consciousness. If it requires an explanation, refactor and add a purpose statment

src/components/treeInfo/index.tsx
```diff
@@ -63,6 +76,25 @@ const TreeInfo: React.FC<TreeProps> = ({
     }
   };
 
+  const shareButton = (
+    <ShareButton
+      size={mobile ? 'middle' : 'large'}
+      // if user owns the tree, show first line
```
rymaju, line 82: Same as above, if it needs and explanation then theres a problem

src/components/shareButton/index.tsx
huang0h: I'm not familiar with FI/FII, could you please explain what it means? Thanks!

chromium-52: FI and FII just mean Fundies 1 and 2

src/components/treeInfo/index.tsx
```diff
@@ -21,6 +22,18 @@ const ToggleTextButton = styled(Button)`
...
+  && {
+    background-color: white;
+    border: 1px solid red;
+    color: red;
+  }
+
+  & :hover {
+    background-color: #fff1f1;
+  }
+`;
```
chromium-52, line 35: To get rid of the `&&`, how about something like this?

```suggestion
const UnadoptButton = styled(Button)`
  & :hover {
    background-color: #fff1f1;
  }
`;
```

and then change `type="primary"` on line 127 to `danger`

chromium-52 (approved):
💯 LGTM!

chromium-52 (approved):
👍 LGTM

Tvcz (approved):
lgtm!

PR #162, opened 2022-11-19
Got rid of no site entries bug on tree page

Tvcz:
Modified the tree page to do some null checks on the first site entry, in case the site in question has no site entries. 

In the case that a site does not have site entries, it is treated as if there is no tree present at that site.

## Why

[ClickUp Ticket](https://app.clickup.com/t/2rw2w3n)

No change for end-user, but now the console is less clogged with errors that don't convey any additional information.

## Screenshots

How console would appear previously:
image

Example of previously erroring site now working:
image

No error in console:
image

## Verification Steps

Checked whether the tree pages for a few sites (ids 3331222, 3331216, 3331217, 3331218) which had no site entries in the dev database `site_entries` table now loaded without any errors in the console.

### DIFF
--- a/src/components/treeInfo/index.tsx
+++ b/src/components/treeInfo/index.tsx
@@ -71,7 +71,7 @@ const TreeInfo: React.FC<TreeProps> = ({
             // Display 'Open Planting Site' if no tree has been planted
             // Otherwise, display the tree's commonName or 'Unknown Species' if no commonName exists
             pageTitle={
-              siteData.entries[0] && siteData.entries[0].treePresent
+              siteData.entries[0]?.treePresent
                 ? siteData.entries[0].commonName
                   ? siteData.entries[0].commonName
                   : 'Unknown Species'
--- a/src/components/treeInfo/index.tsx
+++ b/src/components/treeInfo/index.tsx
@@ -47,7 +47,7 @@ const TreeInfo: React.FC<TreeProps> = ({
   const history = useHistory();
   const location = useLocation<RedirectStateProps>();
 
-  const adopted = siteData.entries[0].adopter !== null;
+  const adopted = siteData.entries[0] && siteData.entries[0].adopter !== null;
 
   const getSiteLocation = (): string => {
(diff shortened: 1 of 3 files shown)


src/components/treeInfo/index.tsx
```diff
@@ -71,7 +71,7 @@ const TreeInfo: React.FC<TreeProps> = ({
             // Display 'Open Planting Site' if no tree has been planted
             // Otherwise, display the tree's commonName or 'Unknown Species' if no commonName exists
             pageTitle={
-              siteData.entries[0].treePresent
+              siteData.entries[0] && siteData.entries[0].treePresent
```
chromium-52, line 74: ⛏️ There's an interesting JS syntax that makes this simpler, called the optional chaining operator. Could we use this here (and in `src/containers/treePage/index.tsx`)? For more info, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

```suggestion
              siteData.entries[0]?.treePresent
```

src/components/treeInfo/index.tsx
```diff
@@ -47,7 +47,7 @@ const TreeInfo: React.FC<TreeProps> = ({
   const history = useHistory();
   const location = useLocation<RedirectStateProps>();
 
-  const adopted = siteData.entries[0].adopter !== null;
+  const adopted = siteData.entries[0] && siteData.entries[0].adopter !== null;
```
chromium-52, line 50: Going off of the optional chaining operator comment below, this line should actually be changed to `!!siteData.entries[0]?.adopter` because of how the optional chaining operator returns `undefined` if the object is `undefined` but I admit this isn't super readable so I think what we have now is good

Double exclamation mark: https://stackoverflow.com/questions/29312123/how-does-the-double-exclamation-work-in-javascript

chromium-52 (commented):
👍 Nice! Just two small nitpicks

chromium-52 (approved):
💯 LGTM!

zhangcry (approved):

PR #175, opened 2022-12-18
Start showing tree icons earlier and increase icon sizes

chromium-52:
## Why

[ClickUp Ticket](https://app.clickup.com/t/27pku6x)

Start showing tree icons at zoom level 15 instead of the current 16.
Increase the size of the icons (especially at higher zoom levels) so that they are more visible to users.

## This PR

Replace the png images for each icon type and size (with set dimensions) with SVG images that can be scaled to any size without them becoming blurry.
The new sizes of the icons at each zoom level are arbitrary and we just looked and picked the sizes of the images that looked the best at each zoom level.

## Screenshots

Screenshots of the map at various zoom levels (lmk if the sizes of the icons can be bigger/smaller at any of the levels!):

 
Level 15:
image (1)

Level 17:
image (2)

Level 18:
image (3)

Level 20:
image (4)

Level 22 (highest zoom level):
image (5)

## Verification Steps

 
Visually verified that the tree icons appear bigger in all zoom levels and that they appear starting at zoom level 15.


src/components/mapComponents/logic/event.ts
```diff
@@ -158,28 +158,24 @@ export function addHandleZoomChange(
...
-          zoomedIn,
-        );
+        setSitesStyle(sitesLayer, visibleSites, zoomLevel, zoomedIn);
         break;
     }
   });
 }
 
 /**
  * Determines the image size given the zoom level and the view.
```
huang0h, line 168: nit: should also change the purpose statement to match the function's signature

src/components/mapComponents/logic/event.ts
```diff
@@ -158,28 +158,24 @@ export function addHandleZoomChange(
...
-    imageSize = 1;
+export function getImageSize(zoomLevel: number): number {
+  if (zoomLevel >= 21) {
+    return 35;
+  } else if (zoomLevel >= 19) {
+    return 20;
+  } else if (zoomLevel >= 18) {
+    return 13;
+  } else {
+    return 5;
```
huang0h, line 179: Just curious, how would scaling the icon size as a function of the zoom level look compared to discrete size intervals?

huang0h (dismissed):
lgtm! 👍

src/components/mapComponents/logic/event.ts
chromium-52: That's an excellent question and it's actually what we did at first. But the function was quadratic (linear wasn't good enough sadly) with seemingly random constants (`0.23 * Math.pow(zoomLevel, 2) - 3.94 * zoomLevel + 12.21` to be exact). Which do you think is better?

huang0h (approved):
lgtm!

src/components/mapComponents/logic/event.ts
Tvcz: I like the idea of a function as well, maybe we could do a regression to find one that fits our existing discrete points?

Tvcz (approved):
lgtm

src/components/mapComponents/logic/event.ts
chromium-52: This is the function we had: `0.23 * Math.pow(zoomLevel, 2) - 3.94 * zoomLevel + 12.21`. We might be able to round the constants to the nearest tenths, do you think this would be better?

huang0h: Fitting our points to a regression could make changing the sizes in the future a lot harder, since whoever's making the change may not know which points it's fit to and will have to play around with the coefficients to make something work. If we want to do a function, would it be possible to save it until the very end for our sake?

PR #177, opened 2022-12-19
Allows users to name trees they adopt

chromium-52:
## Why

[ClickUp Ticket](https://app.clickup.com/t/2b937c2)

Allow users to name trees (specifically, site entries) that they have adopted (and admin users any trees).

## This PR

- Add name site entry route to available routes
- Combine the desktop and mobile styled components in `pagerHeader` into one and move them to `themedComponents`
- Create the `editTreeNameForm` that allows users to (re)name trees, if allowed:
    - Admin users can (re)name any tree
    - Non-admin users can (re)name a tree only if they have adopted the tree
- Create the `treePageHeader` component that displays the tree's name (along with its common name and the site's address)

## Screenshots

 
If the user is an admin or an adopter and the tree has no name:
admin-or-adopter-no-name

If the user is not an admin or an adopter and the tree has no name:
non-adopter-no-name

If the user is an admin or adopter and the tree has a name:
admin-name

(when the admin or adopter clicks on the pencil icon):
admin-name-change

If the user is not an admin or an adopter and the tree has a name:
non-adopter-name

A few screenshots on mobile:
mobile-1

mobile-2

## Verification Steps

 
Visually verified to make sure that the texts/icons/buttons appear as expected in the above cases (in the screenshot).
Changed names of trees to several different values (including empty string to delete its name) and checked the database to make sure that the changes were reflected there as appropriate.


rymaju:
Finally!

Tvcz (approved):
lgtm

zhangcry (approved):

PR #181, opened 2023-01-15
Frontend Capabilities to Delete a Stewardship Activity

SurabhiKeesara:
## Why

[ClickUp Ticket](https://app.clickup.com/t/1dk44y0)

This change allows users to delete their created stewardship activities. It creates a user interface for existing backend capabilities and adds flexibility for incorrectly inputted/extraneous stewardship activities, thus allowing for increased website accuracy.

## This PR

This change involved the creation and styling of a delete icon that is visible only for a given user's own trees and admin. Once users click this button, a modal pops up to confirm their deletion. From there, the stewardship activity is deleted and no longer visible in the list of documented stewardship activities.

## Screenshots

Activity Display Window
Screen Shot 2023-01-21 at 6 48 56 PM

On hover
Screen Shot 2023-01-21 at 6 49 02 PM

Modal
Screen Shot 2023-01-15 at 12 37 18 PM

On hover
Screen Shot 2023-01-15 at 12 37 28 PM

## Verification Steps

1. Using an account on my local environment, I adopted a tree and added a number of various stewardship activities, making sure to include various combinations to ensure that the position of the delete button is good. 
2. From there, I clicked to delete a stewardship activity and ensured that the Modal popped up with all correct information. 
3. I verified that upon selecting "Delete" in the Modal, the Stewardship activity would no longer appear in the list of selected activities. This was also tested for activities at different positions in the list to ensure that no indexing errors were occurring and the correct activity id was passed.
4. I then created an alternative non-admin account using a different email on my local environment and went to the same previously visited tree to ensure that users and non-admin who did not adopt that tree cannot see the delete button. 
5. I then also added a tree and stewardship activities to the secondary account to make sure that my primary admin account can see the delete button.


huang0h (approved):
💯 lgtm!

chromium-52 (changes requested):
Great job! Just had a few quick notes

huang0h:
I tried out the changes on my machine to see how it would fit with editing stewardships, and I ran into a bug where the wrong activities would be deleted:



It seems like only the last one ever gets deleted - not sure if anyone else got this too, or if something's just wrong on my end.

src/components/treeActivity/index.tsx
```diff
@@ -57,6 +63,27 @@ const CenteredPagination = styled(Pagination)`
   }
 `;
 
+const DeleteActivityButton = styled(LinkButton)`
+  color: white;
+  margin: 10px;
+  padding: 0px 10px;
+  background: #d07b69;
```
chromium-52, line 70: 🔧 Could we be able to turn `#d07b69` into a constant (we can call it light red or pink or salmon or something) and define it in `../../utils/colors` (similar to the other color constants)? Same for line 74 as well.

src/components/treeActivity/index.tsx
```diff
@@ -122,6 +205,18 @@ const TreeActivity: React.FC<TreeActivityProps> = ({
...
+                    onClick={() => {
+                      setShowForm(!showForm);
+                      setActivityToDelete(value.activityId);
+                    }}
+                  >
+                    <DeleteOutlined />
+                    {DeleteModalDisplay()}
+                  </DeleteActivityButton>
+                )}
               </Col>
```
chromium-52, line 220: ⛏️ Could we be able to do this so that the text wraps around correctly (just adding a new column for the buttons)?

```suggestion
              <Col span={16}>
                <EntryMessage>{value.message}</EntryMessage>
              </Col>
              <Col span={2}>
                {(value.userId === currentUserId || userIsAdmin) && (
                  <DeleteActivityButton
                    type="primary"
                    onClick={() => {
                      setShowForm(!showForm);
                      setActivityToDelete(value.activityId);
                    }}
                  >
                    <DeleteOutlined />
                    {DeleteModalDisplay()}
                  </DeleteActivityButton>
                )}
              </Col>
```

image

image

src/containers/treePage/ducks/selectors.ts
```diff
@@ -40,6 +42,8 @@ export const mapStewardshipToTreeCare = (
       if (item.watered) activityStrings.push('watered');
       if (item.weeded) activityStrings.push('weeded');
       return {
+        activityId,
```
chromium-52, line 45: 🔧 The [build](https://github.com/Code-4-Community/speak-for-the-trees-frontend/actions/runs/3976899519/jobs/6817668046) is failing because the `TreeCare` interface was changed. We can check this locally by running `npm run check` in the terminal. And this can be fixed by changing lines 54 through 72 in the relevant test file (`src\containers\treePage\test\selectors.test.ts`) to the following (just adding the correct `activityId` and `userId` to the expected output):

```
    it('returns correctly formatted strings with different stewardship combinations', () => {
      const expectedTreeCare: TreeCare[] = [
        {
          activityId: 0,
          userId: 1,
          day: '1st',
          month: 'Jan',
          year: 2021,
          message: 'Was mulched and watered and weeded.',
        },
        {
          activityId: 1,
          userId: 1,
          day: '23rd',
          month: 'Feb',
          year: 2021,
          message: 'Was cleared of waste.',
        },
      ];
```

chromium-52 (commented):
💯 Looks good! Just three quick notes

chromium-52 (approved):
👍 LGTM!

huang0h (approved):
lgtm! 👍

PR #186, opened 2023-01-22
Update Map Pin from Edit Site Form

Tvcz:
## Why

[ClickUp Ticket](https://app.clickup.com/t/2v888qn)

Updates the map pin in real time to match the coordinates inputted into the edit site form, similar to how the form updates when the map pin is moved manually.

## This PR

Adds a marker state to the `sitePage` and `adminDashboard` components so that the state of the marker can be accessed in a callback sent down to the `editSiteForm`, which calls the callback when the latitude or longitude values in the form are modified.

## Verification Steps

Manually checked the admin page (`/admin`) and several edit site features pages (`/site/3215646`, `/site/3310969`, `/site/3321019`) to ensure that changing the coordinates on the form moved the pin and that moving the pin still updated the form.

### DIFF
--- a/src/components/forms/editSiteForm/index.tsx
+++ b/src/components/forms/editSiteForm/index.tsx
@@ -10,16 +10,39 @@ import styled from 'styled-components';
 
 interface EditSiteFormProps {
   readonly formInstance: FormInstance<EditSiteRequest>;
+  readonly onEdit: (formLat: number, formLng: number) => void;
 }
 
 const EditForm = styled(Form)`
   width: 100%;
 `;
 
-const EditSiteForm: React.FC<EditSiteFormProps> = ({ formInstance }) => {
+// function updateMapPin(lat: number, lng: number) {
+//   // If the place does not have a geometry (if the user did not enter a valid location)
+//   const pinLatLng = new google.maps.LatLng(lat, lng);
+//
+//   // Goes to the place they searched for
+//   const pinMarker = new google.maps.Marker();
+//   pinMarker.setPosition(pinLatLng);
+//   zoomToLocation(pinLatLng, map, zoomLevel);
+// }
+
+const EditSiteForm: React.FC<EditSiteFormProps> = ({
+  formInstance,
+  onEdit,
+}) => {
   return (
     <>
-      <EditForm name="basic" form={formInstance}>
+      <EditForm
+        name="basic"
+        form={formInstance}
+        onValuesChange={(changedValues, allValues) => {
+          if (changedValues.lat || changedValues.lng) {
+            // @ts-ignore
+            onEdit(allValues.lat, allValues.lng);
+          }
+        }}
+      >
         <Flex>
           <TitleStack title={'Address'} flexGrow={'1'}>
             <Form.Item
--- a/src/components/forms/editSiteForm/index.tsx
+++ b/src/components/forms/editSiteForm/index.tsx
@@ -17,16 +17,6 @@ const EditForm = styled(Form)`
   width: 100%;
 `;
 
-// function updateMapPin(lat: number, lng: number) {
-//   // If the place does not have a geometry (if the user did not enter a valid location)
-//   const pinLatLng = new google.maps.LatLng(lat, lng);
-//
-//   // Goes to the place they searched for
-//   const pinMarker = new google.maps.Marker();
-//   pinMarker.setPosition(pinLatLng);
-//   zoomToLocation(pinLatLng, map, zoomLevel);
-// }
-
 const EditSiteForm: React.FC<EditSiteFormProps> = ({
   formInstance,
   onEdit,
--- a/src/components/siteFeatures/index.tsx
+++ b/src/components/siteFeatures/index.tsx
@@ -16,12 +16,14 @@ interface SiteFeaturesProps {
   readonly site: SiteProps;
   readonly editSiteForm: FormInstance<EditSiteRequest>;
   readonly onSubmit: (request: EditSiteRequest) => void;
+  readonly onEdit: (formLat: number, formLng: number) => void;
 }
 
 const SiteFeatures: React.FC<SiteFeaturesProps> = ({
   site,
   editSiteForm,
   onSubmit,
+  onEdit,
 }) => {
   const [editingFeatures, setEditingFeatures] = useState(false);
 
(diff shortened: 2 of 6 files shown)


src/components/forms/editSiteForm/index.tsx
```diff
@@ -10,16 +10,29 @@ import styled from 'styled-components';
...
+}) => {
   return (
     <>
-      <EditForm name="basic" form={formInstance}>
+      <EditForm
+        name="basic"
+        form={formInstance}
+        onValuesChange={(changedValues, allValues) => {
+          if (changedValues.lat || changedValues.lng) {
+            // @ts-ignore
```
chromium-52, line 31: ❓Does typescript show an error without this line?

src/components/forms/editSiteForm/index.tsx
```diff
@@ -10,16 +10,29 @@ import styled from 'styled-components';
 
 interface EditSiteFormProps {
   readonly formInstance: FormInstance<EditSiteRequest>;
+  readonly onEdit: (formLat: number, formLng: number) => void;
```
chromium-52, line 13: ⛏️ Just a thought but how about we combine `formLat` and `formLng` into a `google.maps.LatLng` because those two are paired (i.e. the combination of `formLat` and `formLng` has a specific meaning)?

src/components/forms/editSiteForm/index.tsx
```diff
@@ -10,16 +10,29 @@ import styled from 'styled-components';
...
+  formInstance,
+  onEdit,
+}) => {
   return (
     <>
-      <EditForm name="basic" form={formInstance}>
+      <EditForm
+        name="basic"
+        form={formInstance}
+        onValuesChange={(changedValues, allValues) => {
```
chromium-52, line 29: 🔧 Could we move this function outside of the return statement for readability purposes? Sort of like:

```
const EditSiteForm = ({ formInstance, onEdit }) => {
  const onValuesChange = (changedValues, allValues) => {
    ...
  }

  return (
    <>
      <EditSiteForm ... onValuesChange={onValuesChange}>
    ...
    </>
  )
}
```

src/containers/adminDashboard/index.tsx
```diff
@@ -153,7 +155,14 @@ const AdminDashboard: React.FC<AdminDashboardProps> = ({
               <Block
                 maxWidth={windowType === WindowTypes.Mobile ? '100%' : '45%'}
               >
-                <EditSiteForm formInstance={editSiteForm} />
+                <EditSiteForm
+                  formInstance={editSiteForm}
+                  onEdit={(formLat: number, formLng: number) =>
```
chromium-52, line 160: 🔧 Same note as above

src/containers/sitePage/index.tsx
```diff
@@ -108,6 +109,9 @@ const SitePage: React.FC<SitePageProps> = ({ neighborhoods, sites }) => {
                 site={site}
                 editSiteForm={editSiteForm}
                 onSubmit={onSubmitEditSite}
+                onEdit={(formLat: number, formLng: number) =>
```
chromium-52, line 112: 🔧 Same note as above

src/containers/adminDashboard/index.tsx
```diff
@@ -70,6 +70,8 @@ const AdminDashboard: React.FC<AdminDashboardProps> = ({
   const { windowType } = useWindowDimensions();
   const dispatch = useDispatch();
 
+  const [marker, setMarker] = useState<google.maps.Marker>();
```
chromium-52, line 73: ⛏️ Can we rename this state to something like `mapSearchMarker` so that its purpose is slightly more obvious?

src/containers/sitePage/index.tsx
```diff
@@ -51,6 +51,7 @@ interface SitePageProps {
 
 const SitePage: React.FC<SitePageProps> = ({ neighborhoods, sites }) => {
   const [site, setSite] = useState<SiteProps>();
+  const [marker, setMarker] = useState<google.maps.Marker>();
```
chromium-52, line 54: ⛏️ Same note as above

chromium-52 (changes requested):
💯 Awesome! Just a few quick changes and suggestions

src/components/siteFeatures/index.tsx
```diff
@@ -41,7 +43,7 @@ const SiteFeatures: React.FC<SiteFeaturesProps> = ({
     case true:
```
huang0h, line 43: ❓ Do we know why whoever originally wrote this code used a switch statement here? It looks like it could be simplified to an if statement or JavaScript's [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator)

Tvcz: not sure, I guess an argument could be made for readability

src/components/forms/editSiteForm/index.tsx
Tvcz: yeah 😢

chromium-52: I like how I get a warning on that too :(
image

Could we just do this?

```suggestion
            const values = allValues as EditSiteRequest;
            onEdit(new google.maps.LatLng(values.lat, values.lng));
````

File: src/components/siteFeatures/index.tsx (hunk 8 above)
Comment (chromium-52) [reply src/components/siteFeatures/index.tsx:43]: I would even argue using the switch is _less_ readable but 🤷

**Review by chromium-52 — APPROVED** [review]
👍 lgtm!

File: src/components/siteFeatures/index.tsx (hunk 8 above)
Comment (huang0h) [reply src/components/siteFeatures/index.tsx:43]: yeah, at first I thought there was some case for if editingFeatures was undefined, or something else unexpected

**Review by huang0h — APPROVED** [review]
lgtm!

**Review by SurabhiKeesara — DISMISSED** [review]

File: src/components/forms/editSiteForm/index.tsx [hunk 9]
```diff
@@ -10,16 +10,30 @@ import styled from 'styled-components';
...
 const EditForm = styled(Form)`
   width: 100%;
 `;
 
-const EditSiteForm: React.FC<EditSiteFormProps> = ({ formInstance }) => {
+const EditSiteForm: React.FC<EditSiteFormProps> = ({
+  formInstance,
+  onEdit,
+}) => {
+  const onValuesChange = (changedValues: any, allValues: any) => {
```
chromium-52, line 24: 🔥

chromium-52 (approved):
👍 lgtm!!

PR #187, opened 2023-01-23
Default site name input to current name and refresh page on name change

chromium-52:
## Checklist

- [X] 1. Run `npm run check`
- [X] 2. Run `npm run test`

## Why

[ClickUp Ticket](https://app.clickup.com/t/862j171zh)

1. Default the value of site name text input to be the current name so that the user won't have to type the entire site name when changing just a few characters
2. Refresh the page when the user changes the name of a site so that the user sees that the site name was changed without having to manually refresh the page

## This PR

1. Set the default value of the text input in `EditTreeNameForm` to be the current tree name, if applicable
2. Use the `useDispatch` hook to get the updated site data and thus trigger a re-render of the component

## Verification Steps

Visually verified that the default value of the tree name input is the current site name both when the site has and doesn't have a name

Tried changing the site name to arbitrary text (including empty text to delete the site's name) and visually verified that the component re-renders (i.e. the page refreshes)

### DIFF
--- a/src/components/forms/editTreeNameForm/index.tsx
+++ b/src/components/forms/editTreeNameForm/index.tsx
@@ -11,6 +11,7 @@ interface StyledButtonProps {
 }
 
 interface EditTreeNameFormProps extends StyledButtonProps {
+  readonly treeName: string;
   readonly editTreeNameForm: FormInstance<NameSiteEntryRequest>;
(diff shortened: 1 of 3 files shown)


Tvcz (approved):
lgtm

huang0h (approved):
lgtm! 👍

PR #190, opened 2023-01-27
Modify stewardship activity message

huang0h:
## Checklist

- [x] 1. Run `npm run check`
- [x] 2. Run `npm run test`

## Why

(no ClickUp ticket)

Makes the message associated with each stewardship activity less verbose.

## This PR

- Added function `generateTreeCareMessage` in `src/utils/stringFormat.tsx`
- Added tests for `generateTreeCareMessage`
- Modify `mapStewardshipToTreeCare` selector to make use of new function

## Screenshots

Before:

After:

## Verification Steps

- Verified updated messages appear correctly


src/utils/stringFormat.tsx
```diff
@@ -158,3 +162,24 @@ export function parseLatLng(str: string): Coordinate | null {
...
+export function generateTreeCareMessage(item: Activity): string {
+  const activityStrings = [] as string[];
+  if (item.cleaned) activityStrings.push('cleared of waste');
+  if (item.mulched) activityStrings.push('mulched');
+  if (item.watered) activityStrings.push('watered');
+  if (item.weeded) activityStrings.push('weeded');
+
+  const numberOfActivities = activityStrings.length;
+
+  // invariant: at least one activity will be present
```
chromium-52, line 175: ⛏️ Just to be safe, could we add this here? Because technically, there's nothing stopping someone from passing in all `false`s

```
if (numberOfActivities === 0) {
  throw new Error('At least one activity must be true');
}
```

chromium-52 (approved):
👍 Looks great! Haven't looked too closely but there seems to be an interesting function that does this too: https://stackoverflow.com/a/53641525 (just an FYI)

src/utils/stringFormat.tsx
huang0h: `Intl.ListFormat` seems really cool! Would love to use it, but looks like it's really [only supported by browser versions from 2020/1 onwards](https://caniuse.com/?search=Intl.ListFormat) - is that something we really worry about?

chromium-52: I think for our case, 91% global support is good enough. For reference, syntax like the [optional chaining operator](https://caniuse.com/mdn-javascript_operators_optional_chaining) and the [nullish coalescing operator](https://caniuse.com/mdn-javascript_operators_nullish_coalescing) (which we do use sporadically) have similar browser support.

chromium-52:
Looked into the `npm run build` failing and I think we might have to change the `target` in `tsconfig.json` to `es2021` in order for it to work. This is so much cleaner but let's just revert back to what we had for now :(

huang0h:
Yeah, was taking a look too and couldn't find anything either - reverting now :(

chromium-52 (dismissed):
👍 lgtm!

src/utils/stringFormat.tsx
```diff
@@ -176,6 +177,29 @@ export function parseLatLng(str: string): Coordinate | null {
   return [lat, lng];
 }
 
+export function generateTreeCareMessage(item: Activity): string {
+  const activityStrings = [] as string[];
+  if (item.cleaned) activityStrings.push('cleared of waste');
+  if (item.mulched) activityStrings.push('mulched');
+  if (item.watered) activityStrings.push('watered');
+  if (item.weeded) activityStrings.push('weeded');
```
SurabhiKeesara, line 185: Great job! I think that something to look into is marking stewardship activities as icons down the line. For now this formatting is great, but as we add functionality like delete/edit stewardship activities, having icons vs long activity messaging might look cleaner

SurabhiKeesara (approved):
lgtm!

src/utils/stringFormat.tsx
huang0h: Yeah that's a really good idea! We might have to make a couple icons for the more tree-specific activities like mulching and weeding.

PR #191, opened 2023-01-29
Added tree name to share text

Tvcz:
## Checklist

- [x] 1. Run `npm run check`
- [x] 2. Run `npm run test`

## Why

[ClickUp Ticket](https://app.clickup.com/t/863fvut9w)

Allows the user to include tree species in sharing message, makes the message nicer.

## This PR

Small change made to treeInfo page, added a second string interpolation for the sake of readability instead of combining it with the older one, which might have looked like a confusing ternary statement.

## Screenshots

image

## Verification Steps

Manually checked whether the output looked correct on a number of sample trees: (`/tree/3320194`,  `/tree/3329371`, and `/tree/3027901`).


src/components/treeInfo/index.tsx
```diff
@@ -92,7 +92,7 @@ const TreeInfo: React.FC<TreeProps> = ({
           ? 'Check out this tree I adopted!'
           : adopted
           ? 'Check out this tree near you!'
-          : `This tree${
+          : `This ${(siteData.entries[0].commonName ?? 'tree').toLowerCase()}${
```
chromium-52, line 95: 🔧 Could we show the tree name on lines 92 and 94 as well? This will probably require a helper function (preferably with unit tests). Also, we might have to do something like `siteData.entries[0]?.commonName` to account for the case where a site has no site entries

chromium-52 (changes requested):

src/components/treeInfo/index.tsx
SurabhiKeesara: I like the use of 'tree' as a default value, but I agree that maybe you can clean this up by defining some local treeName variable that is either siteData.entries[0].commonName or 'tree' when the site has no entries. And then add that variable to lines 92, 94, and 95

SurabhiKeesara: I like the use of 'tree' as a default value, but I agree that maybe you can clean this up by defining some local treeName variable that is either siteData.entries[0].commonName or 'tree' when the site has no entries. And then add that variable to lines 92, 94, and 95

SurabhiKeesara (commented):
looks good for the most part, just minor cleanup/dealing with the no site entries case

src/components/treeInfo/index.tsx
Tvcz: Addressed feedback!

src/components/treeInfo/index.tsx
```diff
@@ -55,6 +55,9 @@ interface TreeProps {
   readonly onClickEditTreeName: (values: NameSiteEntryRequest) => void;
 }
 
+export const getTreeSpecies = (siteData: SiteProps) =>
```
chromium-52, line 58: ⛏️ Can we change the function's name to something like `getCommonName`? Just because we're technically getting its common name, not species

Also, can we move this function and its tests to the `/utils` folder (maybe `treeFunctions.tsx` and `treeFunctions.test.ts`)?

src/components/treeInfo/treeInfo.test.ts
```diff
@@ -0,0 +1,80 @@
...
+  lat: 324234,
+  lng: 234423,
+  city: 'Boston',
+  zip: '02115',
+  address: '21 Forsyth St.',
+  neighborhoodId: 241142,
+  entries: [],
+};
+
+describe('tests for the getTreeSpecies function', () => {
```
chromium-52, line 69: 🔥

chromium-52 (commented):
💯 Nice! Just two quick comments

src/components/treeInfo/index.tsx
Tvcz: Addressed!

chromium-52 (commented):
💯 Just one nitpick!

chromium-52 (approved):
👍lgtm!!

PR #204, opened 2023-02-28
Add toggling map view functionality on primary map

huang0h:
## Checklist

- [x] 1. Run `npm run check`
- [x] 2. Run `npm run test`

## Why

[ClickUp Ticket](https://app.clickup.com/t/2v888tr)

Allows users to switch between a roadmap view (current view) or a more detailed satellite view of Boston

## This PR

- Add new site icons to fit the satellite map view
- Implement a switch as a custom control to toggle the map view
- Update map styling (setting regions + sites) to change based on map view
- Slightly increased site icon sizes

## Screenshots

Legend + satellite view standard, adopted, and open icons:

a-t1

Satellite view standard, adopted, and young icons:

a-t2

Satellite view neighborhoods:

image

## Verification Steps

- Verified switch properly changes the map view, regions, and icons
- Verified map legend displays proper icons based on map view
- Verified other maps are unaffected and still show roadmap icons


src/components/mapComponents/maps/mapWithPopup/index.tsx
```diff
@@ -189,6 +217,7 @@ const MapWithPopup: React.FC<MapWithPopupProps> = ({
           onChange={(event) => setSearchInput(event.target.value)}
         />
       </div>
+      {toggleViewCard && createPortal(toggleViewCard, toggleViewDiv.current)}
```
huang0h, line 220: A bit of explanation for this:

I wanted to make the toggle view card a [google maps custom control](https://developers.google.com/maps/documentation/javascript/examples/control-custom) so it's automatically positioned nicely as the window is resized (the styling for the given component is relative to the top right corner of the map). However, custom controls must be native DOM elements, not React components, We can attach a dummy div as our control, then use a [react portal](https://reactjs.org/docs/portals.html) to attach a component. 

Unfortunately, since we render the map in a `useEffect` hook, if we create a plain div through `document.createElement`, I believe what happens is the div with the component gets overwritten by an empty div without the component during a re-render, so nothing shows up - we can use a `useRef` hook for a "stateful" div and properly show the card.

tl;dr i don't like useEffect

PR #213, opened 2023-04-04
Connect force unadopt site route

chromium-52:
## Checklist

- [x] 1. Run `npm run check`
- [x] 2. Run `npm run test`

## Why

[ClickUp Ticket](https://app.clickup.com/t/1dk46ne)

Allow admin users to force unadopt a site (even though showing the adopter's email is part of this ticket, I don't think it's worth it given that it can (relatively easily) be done through the reports page)

## This PR

Add a force unadopt site button (for admin users only) that allows admins to force a user to unadopt a site

## Screenshots

image

image

## Verification Steps

Check that the force unadopt button appears only when the users is admin and successfully unadopts the given site, even if the admin is not the site's adopter

### DIFF
--- a/src/containers/treePage/index.tsx
+++ b/src/containers/treePage/index.tsx
@@ -6,7 +6,7 @@ import { Col, Form, message, Row, Typography, Alert } from 'antd';
 import { RedirectStateProps, Routes } from '../../App';
 import { Helmet } from 'react-helmet';
 import { UserAuthenticationReducerState } from '../../auth/ducks/types';
-import { isAdmin, isLoggedIn } from '../../auth/ducks/selectors';
+import { isLoggedIn } from '../../auth/ducks/selectors';
 import {
   ActivityRequest,
   MonthYearOption,
--- a/src/containers/treePage/index.tsx
+++ b/src/containers/treePage/index.tsx
@@ -6,7 +6,7 @@ import { Col, Form, message, Row, Typography, Alert } from 'antd';
 import { RedirectStateProps, Routes } from '../../App';
 import { Helmet } from 'react-helmet';
 import { UserAuthenticationReducerState } from '../../auth/ducks/types';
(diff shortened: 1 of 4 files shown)


src/containers/treePage/index.tsx
```diff
@@ -180,6 +181,19 @@ const TreePage: React.FC<TreeProps> = ({
...
+  const onClickForceUnadopt = () => {
+    protectedApiClient
+      .forceUnadoptSite(id)
+      .then(() => {
+        message.success('Force unadopted site!');
+        dispatch(getSiteData(id));
+        dispatch(getAdoptedSites());
+      })
+      .catch((err) => {
+        message.error(`Failed to force unadopt site: ${err.response.data}`);
```
huang0h, line 193: 💭 Would it be worth our time to do a little refactoring at some point to move some of these functions into TreeInfo or even further down? It seems like a lot of them just rely on the dispatch function from `useDispatch` and the site id from lines 125/126 - maybe we could utilize a React context here to reduce how much prop drilling we do.

huang0h (approved):
Overall lgtm!

src/containers/treePage/index.tsx
chromium-52: I think the idea behind this is that `treePage` is supposed to be a "smart" component, while the components it renders are "dumb" components (more info [here](https://medium.com/@thejasonfile/dumb-components-and-smart-components-e7b33a698d43) and [here](https://javascript.plainenglish.io/react-all-about-components-35650a02ff50) but it's essentially that the page-level component has all of the logic so that all of the components within the page just take in everything and all they have to do is render). This does clutter the "smart" component quite a bit and I can't say I'm fully onboard with this concept but I think at this point, we can just leave it because we do this on p much every page and it'll probably take too long. Great catch tho!

PR #221, opened 2023-04-13
Ah.emailer filters

huang0h:
## Checklist

- [ ] 1. Run `npm run check`
- [ ] 2. Run `npm run test`

## Why

[Ticket 1](https://app.clickup.com/t/863g5624x) and [Ticket 2](https://app.clickup.com/t/863g5624x)

## This PR

- Add new component `emailerFilterControls` to control which users to send emails to
- Implement API method to retrieve common names of all active trees

## Screenshots
2023-04-14 19_02_55-Send Emails — Mozilla Firefox
image
p1

## Verification Steps

- Verified controls correctly adjust filters state


src/api/apiClient.ts
```diff
@@ -107,6 +109,12 @@ const getStewardshipActivities = (
   ).then((res) => res.data);
 };
 
+const getAllCommonNames = (): Promise<string[]> => {
+  return AppAxiosInstance.get(
+    ParameterizedApiRoutes.GET_ALL_COMMON_NAMES(),
+  ).then((res) => res.data.names);
```
chromium-52, line 115: ⛏️ Could we have `getAllCommonNames` return just `res.data` here (instead of `res.data.names`) so that it's consistent with the other routes?

src/containers/email/types.ts
```diff
@@ -0,0 +1,28 @@
+import { Neighborhoods } from '../../assets/content';
+
+export enum EmailType {
+  INACTIVE = 'Inactive',
+  NEIGHBORHOOD = 'Neighborhood',
+}
+
+export interface EmailerFilters {
+  activityCountMin: number;
```
chromium-52, line 9: ❓Should `activityCountMin` also be optional (i.e. `activityCountMin?: number;`)?

src/containers/email/index.tsx
```diff
@@ -1,18 +1,55 @@
...
+  const [filters, setFilters] = useState<EmailerFilters>({
+    activityCountMin: 0,
+    neighborhoods: [],
+    commonNames: [],
+  });
+
+  function generateFilterButtons(key: 'neighborhoods' | 'commonNames') {
+    return filters[key].map((n) => {
+      return (
+        <NeighborhoodFilter
```
chromium-52, line 37: ⛏️ Could we rename this to something like `Filter` just because it's used for both neighborhood and common name filters?

src/components/emailerFilterControls/index.tsx
```diff
@@ -0,0 +1,235 @@
...
+`;
+
+const autoCompleteStyles = { width: 200, marginRight: 15 };
+
+interface EmailerFilterControlsProps {
+  filters: EmailerFilters;
+  setFilters: React.Dispatch<React.SetStateAction<EmailerFilters>>;
+}
+
+const MAX_COUNT = 10;
```
chromium-52, line 33: ❓Can we rename this to something like `MAX_ACTIVITY_COUNT` so that it's a bit more descriptive?

src/components/emailerFilterControls/index.tsx
```diff
@@ -0,0 +1,235 @@
...
+  // Can not select future days
+  return current > moment().endOf('day');
+}
+
+const activityCountLabels: SliderMarks = {
+  0: '0',
+  [MAX_COUNT + 1]: `${MAX_COUNT}+`,
+};
+
+const neighborhoodOptions = Object.values(Neighborhoods)
```
chromium-52, line 56: 🔧 Noticed that the neighborhood names weren't being sorted and I think it's because we're trying to sort an array of objects so we probably need to pass in a custom function for `sort()`. The easier alternative tho I think is just sorting before mapping (i.e. `const neighborhoods = Object.values(Neighborhoods).sort().map((value) => ({ value }));`)

src/components/emailerFilterControls/index.tsx
```diff
@@ -0,0 +1,235 @@
...
+        } else {
+          message.warn(warning);
+        }
+      };
+    },
+    [filters, setFilters],
+  );
+
+  return (
+    <StyledCollapse ghost={true}>
```
chromium-52, line 109: ❓Just curious, what does the `ghost` prop do? Also, just an fyi that if the value of a prop/attribute is `true`, we don't actually need to specify that (in other words, `ghost={true}` can just be `ghost`)

src/components/emailerFilterControls/index.tsx
```diff
@@ -0,0 +1,235 @@
...
+}
+
+function disabledDate(current: moment.Moment): boolean {
+  // Can not select future days
+  return current > moment().endOf('day');
+}
+
+const activityCountLabels: SliderMarks = {
+  0: '0',
+  [MAX_COUNT + 1]: `${MAX_COUNT}+`,
```
chromium-52, line 53: ❓Should `${MAX_COUNT}+` be `${MAX_COUNT + 1}+` (for activity counts, `11+` instead of `10+`)?

src/components/emailerFilterControls/index.tsx
```diff
@@ -0,0 +1,235 @@
...
+          onChange={(_, dateStrings) =>
+            setFilters({
+              ...filters,
+              adoptedStart: dateStrings[0],
+              adoptedEnd: dateStrings[1],
+            })
+          }
+          disabledDate={disabledDate}
+        />
+        <Button
```
chromium-52, line 142: ❓Does the built-in x button in the range picker not clear the start/end dates correctly? If so, we might want to hide this (if it isn't too difficult)

image

src/components/emailerFilterControls/index.tsx
```diff
@@ -0,0 +1,235 @@
...
+              lastActivityStart: undefined,
+              lastActivityEnd: undefined,
+            })
+          }
+        >
+          <CloseOutlined />
+        </Button>
+      </Collapse.Panel>
+      <Collapse.Panel header="Neighborhood" key="neighborhood">
+        <AutoComplete
```
chromium-52, line 184: 🔧 Instead of having the user choose and add the neighborhood or common name (and have it show up below the filters), maybe it might be easier if we use [multiple select](https://ant.design/components/select#components-select-demo-multiple) instead?

chromium-52 (commented):
💯 Really great job on this!! Just a few small comments and potentially using `Select` instead of `Autocomplete` for the neighborhood and common name filters

src/containers/email/types.ts
huang0h: I think since there's a hard lower limit on activity counts at 0, making it an optional prop would just introduce unnecessary complexity - if `activityCountMin` is 0, then that basically means there's no minimum anyways

src/components/emailerFilterControls/index.tsx
huang0h: Ghost makes the collapse panels borderless and transparent, which I think matches the figma design a little better: 

image

src/components/emailerFilterControls/index.tsx
huang0h: I think it's a matter of preference - I personally read 10+ as "more than 10", which is why right now selecting '10' and '10+' are different, but I can also see it being read as "10 or more". Regardless, I'm happy with either way

src/components/emailerFilterControls/index.tsx
huang0h: I didn't even notice that - looks like it's just an [additional boolean flag to hide it](https://ant.design/components/date-picker#common-api)

src/components/emailerFilterControls/index.tsx
huang0h: Yeah that looks way better - I put the filters at the bottom as a placeholder since I think they're supposed to go above the table (which isn't there yet), but having them right in the search bar would probably be easier

chromium-52 (approved):
👍lgtm, really great work on this!!

PR #222, opened 2023-04-17
Fix filter sites route and add send email form

chromium-52:
## Checklist

- [ ] 1. Run `npm run check`
- [ ] 2. Run `npm run test`

## Why

[ClickUp Ticket](https://app.clickup.com/t/xxxxxxx)


src/api/protectedApiClient.ts
```diff
@@ -117,6 +119,15 @@ export interface ProtectedApiClient {
...
   readonly addSites: (request: AddSitesRequest) => Promise<void>;
+  readonly sendEmail: (request: SendEmailRequest) => Promise<void>;
+  readonly filterSites: (
+    treeCommonNames: string[] | null,
+    adoptedStart: Date | null,
+    adoptedEnd: Date | null,
+    lastActivityStart: Date | null,
+    lastActivityEnd: Date | null,
+    neighborhoodIds: number[] | null,
+  ) => Promise<FilterSitesResponse>;
```
huang0h, line 130: Can we pull these fields out into its own type so it's easier to change the interface declaration, api route, and function implementation (maybe call it `FilterSitesRequest` or `FilterSitesParams`)? Also, could we maybe change the `Date` fields to just `string`, since that's how they're stored in the emailer filters state?

src/api/protectedApiClient.ts
```diff
@@ -506,6 +533,32 @@ const addSites = (request: AddSitesRequest): Promise<void> => {
...
+};
+
+const filterSites = (
+  treeCommonNames: string[] | null,
+  adoptedStart: Date | null,
+  adoptedEnd: Date | null,
+  lastActivityStart: Date | null,
+  lastActivityEnd: Date | null,
+  neighborhoodIds: number[] | null,
+): Promise<FilterSitesResponse> => {
```
huang0h, line 549: Would it better for the return type to be `Promise<{ filteredSites: FilteredSite[] }>`? IMO it feels a little unnecessary to create a new type for an object with one field.

huang0h (commented):
lgtm! Just have a couple comments about the filter sites route

src/api/protectedApiClient.ts
chromium-52: Overlooked that I can just squish all of the args into a single object, good call!

src/api/protectedApiClient.ts
chromium-52: I think it's ok because like you mentioned in the comment above, if we were to change its type (like if we wanted to change the name `filteredSites` to something else), we'll only have to change in one place

huang0h:
lgtm!

huang0h (approved):
lgtm!

PR #223, opened 2023-04-18
Ah.emailer table

huang0h:
## Checklist

- [x] 1. Run `npm run check`
- [x] 2. Run `npm run test`

## Why

[ClickUp Ticket](https://app.clickup.com/t/863g5625e)

Adds the table of user info so admins can send emails to them

## This PR

- Implement the table to display user info retrieved from the filter sites route

## Screenshots

image

image

## Verification Steps

- Verified the filter sites data is correctly displayed in the table
- Verified the filter controls correctly affect the data shown in the table


src/containers/email/index.tsx
```diff
@@ -1,41 +1,101 @@
...
+  justify-content: space-between;
+  margin-top: 15px;
+`;
+
+const FetchInfoContainer = styled.div`
+  text-align: center;
+  padding: 30px;
+`;
+
+const selectStyles = { marginTop: 20, minWidth: 150, marginBottom: 20 };
```
chromium-52, line 37: ❓Just curious, what's the reason for doing this instead of using styled components?

src/containers/email/index.tsx
```diff
@@ -1,41 +1,101 @@
...
+  commonNames: [],
+};
+
+enum LoadingState {
+  SUCCESS = 'success',
+  LOADING = 'loading',
+  ERROR = 'error',
+}
+
+function neighborhoodToId(neighborhood: Neighborhoods): number {
```
chromium-52, line 51: ⛏️ Can we move this to `src/utils/stringFormat.tsx` (maybe right below `getNeighborhoodName()`)?

src/containers/email/index.tsx
```diff
@@ -1,41 +1,101 @@
...
-  //   return protectedApiClient
-  //     .filterSites(null, null, null, null, null, neighborhoodIds)
-  //     .then((res: FilterSitesResponse) => console.log(res.filteredSites));
-  // };
+    protectedApiClient
+      .filterSites(req)
+      .then((res) => {
+        setFetchSitesState(LoadingState.SUCCESS);
+        setFetchData(
+          res.filteredSites.filter(
```
chromium-52, line 86: ❓Are the min/max activity count filters not applied to the filtered sites that are returned from the route? If so, I think we should look into this in the backend

src/containers/email/index.tsx
```diff
@@ -56,6 +116,62 @@ const Email: React.FC = () => {
...
+              />
+              <br />
+              <Button type="primary" size="large" onClick={onClickSearch}>
+                Search
+              </Button>
+            </Col>
+            <Col span={1}></Col>
+            <Col span={17}>
+              {(() => {
+                switch (fetchSitesState) {
```
chromium-52, line 145: 💯

chromium-52 (commented):
💯 This is so cool! A couple comments but looks really good overall!

chromium-52:
Just two really small things I saw while playing around with this:
1. For the activity count filter, setting both the min and max activity counts to 0 (to get all of the sites that don't have any activities) makes it so that the max activity count automatically changes to 10+
2. Coming back to the date range pickers in the adoption and last activity date filters, I think the built-in clear button updates the values correctly so we might not need to implement our own clear buttons next to it

src/components/adoptedSitesTable/index.tsx
```diff
@@ -1,102 +1,100 @@
...
-//     name: 'Joe Black',
-//     age: 32,
-//     address: 'Sydney No. 1 Lake Park',
-//     tags: ['cool', 'teacher'],
-//   },
-// ];
+  const [selectedRowKeys, setSelectedRowKeys] = useState<number[]>([]);
 
-// const App: React.FC = () => <Table columns={columns} dataSource={data} />;
+  return fetchData.length > 0 ? (
```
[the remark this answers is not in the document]
huang0h: Yeah, I was between text and an empty table too - I think that's a good case for the empty table

src/containers/email/index.tsx
huang0h: For some reason, when I used a styled component, typescript would keep raising errors with the provided `onChange` function - iirc it's because `Select` has a parameterized value type, and as styled component that value type became unknown so it'd raise an error.

src/components/emailerFilterControls/index.tsx
```diff
@@ -103,12 +99,13 @@ const EmailerFilterControls: React.FC<EmailerFilterControlsProps> = ({
...
         <StyledRangePicker
           allowClear={false}
+          allowEmpty={[true, true]}
           value={formatDates(filters.adoptedStart, filters.adoptedEnd)}
           onChange={(_, dateStrings) =>
             setFilters({
               ...filters,
-              adoptedStart: dateStrings[0],
-              adoptedEnd: dateStrings[1],
+              adoptedStart: dateStrings[0] || undefined,
```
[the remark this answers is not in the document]
huang0h: I don't think it'll break since the range picker accepts null values in each date anyways, so it should fit nicer with the rest of the logic

src/containers/email/index.tsx
huang0h: Yeah, I think it makes more sense to add the activity counts as a parameter in the backend route.

rymaju:
Watching this PR 👀 @huang0h Let me know when you get this approved and merged.

src/containers/email/index.tsx
chromium-52: Can you see if this works?
```
import { SelectProps } from 'antd/lib/select';

...

const EmailTypeSelect = styled((props: SelectProps) => <Select {...props} />)`
  margin-top: 20px;
  min-width: 150px;
  margin-bottom: 20px;
`;
```

and using `<EmailTypeSelect ... />`?

chromium-52 (commented):
💯 This looks good to me, except one small thing I noticed is that in the frontend, an `activityCountMin` (and `activityCountMax`) is `null` if and only if its value is "10+" whereas in the backend, `activityCountMin` (and `activityCountMax`) is `null` if and only if the `activityCountMin` (or `activityCountMax`) filter is not applied. Usually, this actually turns out to be ok unless both `activityCountMin` and `activityCountMax` are set to "10+", which actually returns all adopted sites instead of just sites with more than 10 activities (as shown in the screenshot below). I think we should make the meaning of `activityCountMin` and `activityCountMax` being `null` consistent across frontend and backend

image

src/api/protectedApiClient.ts
```diff
@@ -199,7 +202,7 @@ export const ParameterizedAdminApiRoutes = {
   GET_STEWARDSHIP_REPORT_CSV: (previousDays: number): string =>
     `/api/v1/protected/report/csv/stewardship?previousDays=${previousDays}`,
   FILTER_SITES: (params: FilterSitesParams): string =>
-    `${baseSiteRoute}filter_sites?${
+    `${baseSiteRoute}filter_sites?&activityCountMin=${params.activityCountMin}${
```
chromium-52, line 205: ⛏️the `&` in front of `activityCountMin` isn't necessary

src/utils/stringFormat.tsx
```diff
@@ -213,6 +213,21 @@ export function generateTreeCareMessage(item: Activity): string {
   }
 }
 
+// convert emailer filter values to a text representation of the range
+export function formatActivityCountRange(
+  min: number | null,
+  max: number | null,
+  maxCount: number,
+): string {
+  if (min === null && max === null) {
```
chromium-52, line 222: 🔧 Is it still possible for the first `if` condition to be true?

chromium-52 (commented):
💯 Looks good! Just two really quick comments

src/utils/stringFormat.tsx
huang0h: Nope - thanks for catching this!

chromium-52 (approved):
👍 lgtm, great work!!

PR #232, opened 2023-05-22
Ah.admin redesign

huang0h:
## Checklist

- [ ] 1. Run `npm run check`
- [ ] 2. Run `npm run test`

## Why

Redesign the admin dashboard to consolidate admin features in one page

## This PR

- Refactors the add site forms into a new page
- Move promote user/child account form into modals within buttons
- Add links to admin features under dashboard

## Screenshots

Dashboard: 

image

Promote user modal:

image

Child account modal:

image

### DIFF
--- a/src/containers/adminDashboard/index.tsx
+++ b/src/containers/adminDashboard/index.tsx
@@ -83,11 +83,6 @@ const ImageLinkCard: React.FC<ImageLinkCardProps> = ({
 
 const ICON_SIZE = 40;
 
-interface AdminDashboardProps {
-  readonly neighborhoods: MapGeoDataReducerState['neighborhoodGeoData'];
-  readonly sites: MapGeoDataReducerState['siteGeoData'];
-}
-
 const AdminDashboard: React.FC = () => {
   const privilegeLevel: PrivilegeLevel = useSelector((state: C4CState) =>
     getPrivilegeLevel(state.authenticationState.tokens),
--- a/src/containers/adminDashboard/index.tsx
+++ b/src/containers/adminDashboard/index.tsx
@@ -1,63 +1,81 @@
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import { Helmet } from 'react-helmet';
-import { Form, message, Row, Typography, Divider } from 'antd';
+import { Form, message, Typography, Divider, Modal, Button } from 'antd';
 import PageHeader from '../../components/pageHeader';
 import PageLayout from '../../components/pageLayout';
 import styled from 'styled-components';
-import { connect, useDispatch, useSelector } from 'react-redux';
+import { useSelector } from 'react-redux';
 import { C4CState } from '../../store';
-import { getPrivilegeLevel } from '../../auth/ducks/selectors';
+import {
+  getPrivilegeLevel,
+  getUserFirstName,
+} from '../../auth/ducks/selectors';
+import { LIGHT_GREEN, MID_GREEN } from '../../utils/colors';
 import { PrivilegeLevel } from '../../auth/ducks/types';
 import ChangePrivilegeForm from '../../components/forms/changePrivilegeForm';
-import UploadSitesForm from '../../components/forms/uploadSitesForm';
-import { DARK_GREEN } from '../../utils/colors';
-import ProtectedClient from '../../api/protectedApiClient';
-import {
-  AddSiteRequest,
-  UpdateSiteRequest,
-} from '../../components/forms/ducks/types';
-import UpdateSiteForm from '../../components/forms/updateSiteForm';
-import EditSiteForm from '../../components/forms/editSiteForm';
-import SelectorMapDisplay from '../../components/mapComponents/mapDisplays/selectorMapDisplay';
 import { MapGeoDataReducerState } from '../../components/mapComponents/ducks/types';
-import { MapContainer, Block, Flex } from '../../components/themedComponents';
-import useWindowDimensions, {
-  WindowTypes,
-} from '../../components/windowDimensions';
-import { getMapGeoData } from '../../components/mapComponents/ducks/thunks';
+import { Flex } from '../../components/themedComponents';
 import SignupForm from '../../components/forms/signupForm';
 import { SignupFormValues } from '../../components/forms/ducks/types';
 import ProtectedApiClient from '../../api/protectedApiClient';
 import { AppError } from '../../auth/axios';
 import { getErrorMessage } from '../../utils/stringFormat';
 import { SubmitButton } from '../../components/themedComponents';
-import { round } from 'lodash';
-import { LAT_LNG_PRECISION } from '../../components/forms/constants';
-import { MapTypes } from '../../context/types';
-import { MapTypeContext } from '../../context/mapTypeContext';
+import {
+  BarChartOutlined,
+  FileAddOutlined,
+  MailOutlined,
+  PlusOutlined,
+  RocketFilled,
+  SettingFilled,
+} from '@ant-design/icons';
+import Image2 from '../../assets/images/bkg2.png';
+import Image3 from '../../assets/images/bkg3.png';
+import Image4 from '../../assets/images/bkg4.png';
+import { Routes } from '../../App';
+import Icon, {
+  IconComponentProps,
+} from '@ant-design/icons/lib/components/Icon';
 
 const AdminContentContainer = styled.div`
   width: 80vw;
   margin: 8vh auto auto;
 `;
 
-const DashboardContent = styled.div`
-  width: 450px;
+const ImageCard = styled.div`
+  background: rgba(0, 0, 0, 0.25) url(${(props) => props.background});
+  background-blend-mode: darken;
+  border-radius: 10px;
+  width: 230px;
+  height: 230px;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  color: white;
+  font-size: 16pt;
+  font-weight: bold;
+  transition: background-color 0.2s;
+
+  &:hover {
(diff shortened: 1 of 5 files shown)


src/containers/adminDashboard/index.tsx
```diff
@@ -136,78 +129,72 @@ const AdminDashboard: React.FC<AdminDashboardProps> = ({
...
-            </Flex>
-            <UpdateSiteForm
-              formInstance={updateSiteForm}
-              onFinish={onSubmitAddSite}
-            />
-          </MarginBottomRow>
+          <Typography.Title level={3}>Admin Functions</Typography.Title>
+          <Flex gap={'40px 40px'} margin={'30px 0'}>
+            <ImageLinkCard href={Routes.ADD_SITES} image={Image1}>
+              {/* <LargeIcon component={FileAddOutlined React.ForwardRefExoticComponent<any></any>} /> */}
```
chromium-52, line 159: ⛏️ Is this needed?

src/containers/adminDashboard/index.tsx
```diff
@@ -136,78 +129,72 @@ const AdminDashboard: React.FC<AdminDashboardProps> = ({
...
-                    Create Child Account
-                  </SubmitButton>
-                </Form.Item>
-              </SignupForm>
-            </DashboardContent>
-            <DashboardContent>
-              <Typography.Title level={4}>Add Sites</Typography.Title>
-              <UploadSitesForm />
-            </DashboardContent>
+          <PageHeader
```
chromium-52, line 132: 💯

chromium-52 (dismissed):
👍 lgtm!

CerberusLatrans:
https://github.com/Code-4-Community/speak-for-the-trees-frontend/blob/6d989c4ad0953a217117f9571c3c5767ef7c8dab/src/containers/addSites/index.tsx#LL87C11-L87C11

This is defo just from my lack of understanding of the frontend, but why is there both an editsiteform and updatesiteform? I see that the request made upon `UpdateSiteForm` submission takes values from the `editSiteForm` except for `plantingDate` which comes from `updateSiteForm` -- are these all the values that cannot be changed via the UI?

Other than that, everything else is pretty clear to me. Great job on the redesign!

src/containers/adminDashboard/index.tsx
huang0h: 🤦

huang0h:
> https://github.com/Code-4-Community/speak-for-the-trees-frontend/blob/6d989c4ad0953a217117f9571c3c5767ef7c8dab/src/containers/addSites/index.tsx#LL87C11-L87C11
> 
> This is defo just from my lack of understanding of the frontend, but why is there both an editsiteform and updatesiteform? I see that the request made upon `UpdateSiteForm` submission takes values from the `editSiteForm` except for `plantingDate` which comes from `updateSiteForm` -- are these all the values that cannot be changed via the UI?
> 
> Other than that, everything else is pretty clear to me. Great job on the redesign!

Great questions! For the edit/update site forms, we have the ["sites page"](https://github.com/Code-4-Community/speak-for-the-trees-frontend/blob/master/src/containers/sitePage/index.tsx) (you can access it at [base url]/site/[site id]), which is used to maintain site information. There's a difference between a site and a site entry - a site is the physical location of where a tree may go, so site-specific information includes address, latitude/longitude, etc. A site entry refers to what's actually at the site, namely the tree (if there is one) and information about it, such as species and status. If someone decides to replace the tree at a given site, we'd want to change the site entry information e.g. the species of the new tree, but since the site didn't change - it's still at the same place - we'd like to leave that be. The `EditSiteForm` is used to update site information, while the `UpdateSiteForm` updates site entry information (tbh they could be named better).
tl;dr we have two different forms because we're dealing with two different but related entities 

As for the request, we actually do use all of the information from the update site form - note the ['...request,'](https://github.com/Code-4-Community/speak-for-the-trees-frontend/blob/6d989c4ad0953a217117f9571c3c5767ef7c8dab/src/containers/addSites/index.tsx#LL86C1-L86C1) where we define the `addSiteRequest`. This is an example of JavaScript's [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) - in this case, we're using the spread syntax to take all of the key/value pairs from `request` and copying them into our new variable `addSiteRequest`. Finally, since we want to format the planting date in a specific way, we explicitly set the `plantingDate` again.

Hopefully this all makes sense - let me know if have any more questions!

CerberusLatrans:
> > https://github.com/Code-4-Community/speak-for-the-trees-frontend/blob/6d989c4ad0953a217117f9571c3c5767ef7c8dab/src/containers/addSites/index.tsx#LL87C11-L87C11
> > This is defo just from my lack of understanding of the frontend, but why is there both an editsiteform and updatesiteform? I see that the request made upon `UpdateSiteForm` submission takes values from the `editSiteForm` except for `plantingDate` which comes from `updateSiteForm` -- are these all the values that cannot be changed via the UI?
> > Other than that, everything else is pretty clear to me. Great job on the redesign!
> 
> Great questions! For the edit/update site forms, we have the ["sites page"](https://github.com/Code-4-Community/speak-for-the-trees-frontend/blob/master/src/containers/sitePage/index.tsx) (you can access it at [base url]/site/[site id]), which is used to maintain site information. There's a difference between a site and a site entry - a site is the physical location of where a tree may go, so site-specific information includes address, latitude/longitude, etc. A site entry refers to what's actually at the site, namely the tree (if there is one) and information about it, such as species and status. If someone decides to replace the tree at a given site, we'd want to change the site entry information e.g. the species of the new tree, but since the site didn't change - it's still at the same place - we'd like to leave that be. The `EditSiteForm` is used to update site information, while the `UpdateSiteForm` updates site entry information (tbh they could be named better). tl;dr we have two different forms because we're dealing with two different but related entities
> 
> As for the request, we actually do use all of the information from the update site form - note the ['...request,'](https://github.com/Code-4-Community/speak-for-the-trees-frontend/blob/6d989c4ad0953a217117f9571c3c5767ef7c8dab/src/containers/addSites/index.tsx#LL86C1-L86C1) where we define the `addSiteRequest`. This is an example of JavaScript's [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) - in this case, we're using the spread syntax to take all of the key/value pairs from `request` and copying them into our new variable `addSiteRequest`. Finally, since we want to format the planting date in a specific way, we explicitly set the `plantingDate` again.
> 
> Hopefully this all makes sense - let me know if have any more questions!

That makes a lot more sense-- thanks for the thorough explanation. Just to be sure-- would it be correct to say that the `onSubmitAddSite`, though technically handling the `UpdateSiteForm` submission, is also submitting the values inputted in the `EditeSiteForm` (by retrieving them indirectly)?

src/containers/adminDashboard/index.tsx
```diff
@@ -1,84 +1,107 @@
...
+  <a href={href}>
+    <ImageCard style={{ backgroundImage: `url(${image})` }}>
+      {children}
+    </ImageCard>
+  </a>
+);
+
+const ICON_SIZE = 40;
 
 interface AdminDashboardProps {
```
chromium-52, line 86: ⛏️ this should be an unused interface that can be deleted (not sure why the linter didn't catch this tho)

chromium-52 (approved):
👍 Quick comment but lgtm!

huang0h:
> > > https://github.com/Code-4-Community/speak-for-the-trees-frontend/blob/6d989c4ad0953a217117f9571c3c5767ef7c8dab/src/containers/addSites/index.tsx#LL87C11-L87C11
> > > This is defo just from my lack of understanding of the frontend, but why is there both an editsiteform and updatesiteform? I see that the request made upon `UpdateSiteForm` submission takes values from the `editSiteForm` except for `plantingDate` which comes from `updateSiteForm` -- are these all the values that cannot be changed via the UI?
> > > Other than that, everything else is pretty clear to me. Great job on the redesign!
> > 
> > 
> > Great questions! For the edit/update site forms, we have the ["sites page"](https://github.com/Code-4-Community/speak-for-the-trees-frontend/blob/master/src/containers/sitePage/index.tsx) (you can access it at [base url]/site/[site id]), which is used to maintain site information. There's a difference between a site and a site entry - a site is the physical location of where a tree may go, so site-specific information includes address, latitude/longitude, etc. A site entry refers to what's actually at the site, namely the tree (if there is one) and information about it, such as species and status. If someone decides to replace the tree at a given site, we'd want to change the site entry information e.g. the species of the new tree, but since the site didn't change - it's still at the same place - we'd like to leave that be. The `EditSiteForm` is used to update site information, while the `UpdateSiteForm` updates site entry information (tbh they could be named better). tl;dr we have two different forms because we're dealing with two different but related entities
> > As for the request, we actually do use all of the information from the update site form - note the ['...request,'](https://github.com/Code-4-Community/speak-for-the-trees-frontend/blob/6d989c4ad0953a217117f9571c3c5767ef7c8dab/src/containers/addSites/index.tsx#LL86C1-L86C1) where we define the `addSiteRequest`. This is an example of JavaScript's [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) - in this case, we're using the spread syntax to take all of the key/value pairs from `request` and copying them into our new variable `addSiteRequest`. Finally, since we want to format the planting date in a specific way, we explicitly set the `plantingDate` again.
> > Hopefully this all makes sense - let me know if have any more questions!
> 
> That makes a lot more sense-- thanks for the thorough explanation. Just to be sure-- would it be correct to say that the `onSubmitAddSite`, though technically handling the `UpdateSiteForm` submission, is also submitting the values inputted in the `EditeSiteForm` (by retrieving them indirectly)?

Yep, I'd say that's a good way of thinking about it!

PR #258, opened 2023-06-29
Ah.prevent open adoption

huang0h:
## Checklist

- [ ] 1. Run `npm run check`
- [ ] 2. Run `npm run test`
- [ ] 3. Run `npm build --if-present`  

## Why

See https://github.com/Code-4-Community/speak-for-the-trees-backend-v2/pull/196

Hide adopt buttons and stewardship form if a site does not have a tree present

## Screenshots

image

## Verification Steps

- Verified the adopt buttons/stewardship form don't show if there is no tree present at the site
- Verified above components do show if there is a tree present

### DIFF
--- a/src/components/treeInfo/index.tsx
+++ b/src/components/treeInfo/index.tsx
@@ -180,7 +180,7 @@ const TreeInfo: React.FC<TreeProps> = ({
 
           {shareButton}
 
-          {userOwnsTree && (
+          {userOwnsTree && treePresent && (
             <StewardshipContainer>
               <Typography.Title level={3}>
                 {t('actions.record_activity')}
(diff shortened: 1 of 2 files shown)


src/components/treeInfo/index.tsx
```diff
@@ -143,39 +144,43 @@ const TreeInfo: React.FC<TreeProps> = ({
...
-              type="primary"
-              size={mobile ? 'middle' : 'large'}
-              onClick={onClickAdopt}
-              disabled={isAdopted}
-            >
-              {isAdopted ? t('actions.adopted') : t('actions.adopt')}
-            </Button>
-          )}
-
-          {userIsAdmin && (
```
chromium-52, line 165: As a follow-up to this follow-up, I think when a new site entry is added to a site or an existing latest site entry is modified 
where the entry's `treePresent` value is false and the site is already adopted, we should automatically unadopt the site

chromium-52 (dismissed):
👍lgtm! Just one quick follow-up :)

chromium-52 (approved):
👍 Nice!

PR #279, opened 2023-09-19
temp fix for axios

huang0h:
## Checklist

- [ ] 1. Run `yarn run check`
- [ ] 2. Run `yarn run test`

## Why

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -35,82 +35,7 @@ import history from './history';
 import FAQ from './containers/faq';
 import Email from './containers/email';
 import AddSites from './containers/addSites';
-
-/* TEMPORORY AXIOS FIX */
-import { AxiosError, AxiosResponse } from 'axios';
-import store, { LOCALSTORAGE_STATE_KEY } from './store';
-import { asyncRequestIsComplete } from './utils/asyncRequest';
-import { isTokenValid } from './auth/ducks/selectors';
-import { refresh } from './auth/ducks/thunks';
-import { AppAxiosInstance } from './api/apiClient';
-import authClient from './auth/authClient';
-import protectedApiClient from './api/protectedApiClient';
-
-const userAuthenticationExtraArgs = {
-  authClient,
-  protectedApiClient,
-};
-
-const INVALID_ACCESS_TOKEN = 'Given access token is expired or invalid';
-
-// TODO Use throughout application
-export interface AppError extends Omit<AxiosError, 'response'> {
-  readonly response: AxiosResponse<string>;
-}
-
-export const responseErrorInterceptor = async (
-  error: AxiosError,
-): Promise<AxiosResponse> => {
-  const originalRequest = {
-    ...error.config,
-    _retry: true,
-  };
-
-  const tokens: UserAuthenticationReducerState['tokens'] =
-    store.getState().authenticationState.tokens;
-
-  if (
-    asyncRequestIsComplete(tokens) &&
-    error?.response?.status === 401 &&
-    error?.response?.data === INVALID_ACCESS_TOKEN &&
-    isTokenValid(tokens.result.refreshToken) &&
-    !(error.config as any)?._retry
-  ) {
-    await refresh(tokens.result.refreshToken)(
-      store.dispatch,
-      store.getState,
-      userAuthenticationExtraArgs,
-    );
-    return AppAxiosInstance(originalRequest);
-  }
-  if (
-    asyncRequestIsComplete(tokens) &&
-    error?.response?.status === 401 &&
-    error?.response?.data === INVALID_ACCESS_TOKEN &&
-    !isTokenValid(tokens.result.refreshToken)
-  ) {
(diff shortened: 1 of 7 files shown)


chromium-52 (dismissed):
:+1: I wonder if the only reason why the bug is happening is just because the code that adds the request and response interceptors aren't being run at all, could something like this work?

```tsx
/* in `src/App.tsx` */
const App: React.FC = () => {
  useEffect(() => addAxiosInterceptors(), []);
}

/* in `src/auth/axios.ts` */
// lines 40 through 97 of `src/App.tsx` in this PR

export const addAxiosInterceptors = () => {
  AppAxiosInstance.interceptors.response.use( ... );

  AppAxiosInstance.interceptors.request.use((config) => { ... });
}
```

Still p ugly but at least we're all learning something :shrug: (don't make auth more complicated than it needs to be)

huang0h:
> 👍 I wonder if the only reason why the bug is happening is just because the code that adds the request and response interceptors aren't being run at all, could something like this work?
> 
> ```tsx
> /* in `src/App.tsx` */
> const App: React.FC = () => {
>   useEffect(() => addAxiosInterceptors(), []);
> }
> 
> /* in `src/auth/axios.ts` */
> // lines 40 through 97 of `src/App.tsx` in this PR
> 
> export const addAxiosInterceptors = () => {
>   AppAxiosInstance.interceptors.response.use( ... );
> 
>   AppAxiosInstance.interceptors.request.use((config) => { ... });
> }
> ```
> 
> Still p ugly but at least we're all learning something 🤷 (don't make auth more complicated than it needs to be)

This only worked 100% on my end if I called `addAxiosInterceptors()` outside of the component function - when I had it in `useEffect`, the reports page still gave the same error. For some reason, only the calls to get reports didn't have `X-Access-Token` even though other admin calls like emailer searching, force unadopting, and adding sites had the header and worked totally fine, and I have no idea why.

chromium-52:
> This only worked 100% on my end if I called `addAxiosInterceptors()` outside of the component function - when I had it in `useEffect`, the reports page still gave the same error. For some reason, only the calls to get reports didn't have `X-Access-Token` even though other admin calls like emailer searching, force unadopting, and adding sites had the header and worked totally fine, and I have no idea why.

Ah I think this could be because `useEffect`s get run after a render. This means that when we have `addAxiosInterceptors()` in a `useEffect()` in `App.tsx` and we render the `App.tsx` component, we'll first render everything that gets returned by it first and only after that run the `useEffect()`. This isn't an issue for calls like emailer searching and force unadopting because those calls get made after the page gets rendered. But when we go to the reports page, we can see [here](https://github.com/Code-4-Community/speak-for-the-trees-frontend/blob/fdc00a3b9b5e85cb4146fd8d4802de1a38c21628/src/containers/reports/index.tsx#L54-L73) that we make requests to the backend in the page's `useEffect()`. Sadly, as we can see in the console logs in the screenshot below (link to [sandbox](https://codesandbox.io/s/busy-euler-ky8mzc?file=/src/App.jsx) just as an fyi), `useEffect`s in child components get run before those in the parent components and so we are adding the interceptors after we make the requests to get the adoption and stewardship info and getting an error

image

With this being said idk what the best way to get around this would be so I'd say let's just merge this and spend an hour or so to see if we can move the `addAxiosInterceptors()` call to somewhere outside of `App.tsx`. If we can't, then we can just leave this as is and learn to not implement auth like this in the future

chromium-52 (approved):

huang0h:
> > This only worked 100% on my end if I called `addAxiosInterceptors()` outside of the component function - when I had it in `useEffect`, the reports page still gave the same error. For some reason, only the calls to get reports didn't have `X-Access-Token` even though other admin calls like emailer searching, force unadopting, and adding sites had the header and worked totally fine, and I have no idea why.
> 
> 
> 
> Ah I think this could be because `useEffect`s get run after a render. This means that when we have `addAxiosInterceptors()` in a `useEffect()` in `App.tsx` and we render the `App.tsx` component, we'll first render everything that gets returned by it first and only after that run the `useEffect()`. This isn't an issue for calls like emailer searching and force unadopting because those calls get made after the page gets rendered. But when we go to the reports page, we can see [here](https://github.com/Code-4-Community/speak-for-the-trees-frontend/blob/fdc00a3b9b5e85cb4146fd8d4802de1a38c21628/src/containers/reports/index.tsx#L54-L73) that we make requests to the backend in the page's `useEffect()`. Sadly, as we can see in the console logs in the screenshot below (link to [sandbox](https://codesandbox.io/s/busy-euler-ky8mzc?file=/src/App.jsx) just as an fyi), `useEffect`s in child components get run before those in the parent components and so we are adding the interceptors after we make the requests to get the adoption and stewardship info and getting an error
> 
> 
> 
> image
> 
> 
> 
> With this being said idk what the best way to get around this would be so I'd say let's just merge this and spend an hour or so to see if we can move the `addAxiosInterceptors()` call to somewhere outside of `App.tsx`. If we can't, then we can just leave this as is and learn to not implement auth like this in the future
> 
> 

Ah that makes sense, thanks for the explanation!

PR #303, opened 2023-11-12
Site Image Deletion

### LINKED ISSUE #288: Add delete image button on tree sites
**Why**
- allow users to delete their **own** uploaded images for the tree site/entry
- allow admins to delete any uploaded images

**What to do**
- Add delete image route to the API client
- Add the ability for any user to delete their own uploaded images for a site

Design idea: delete button on the site page that adds a pop-up for confirmation of delete image (similar modal issues)

>Use thumbnails from events in llb as a blueprint (https://github.com/Code-4-Community/lucys-love-bus-frontend/blob/a87c610ad76ba098c1c6b74f76d2f77566a7c377/src/containers/editEvent/index.tsx)
> AntD delete design: https://ant.design/docs/spec/data-list#delete-directly
> AntD modal component: https://ant.design/components/modal

SurabhiKeesara:
## Checklist

- [X] 1. Run `yarn run check`
- [X] 2. Run `yarn run test`

## Why

Resolves #288 

This ticket adds the delete button to site images, allowing users to delete uploaded images. This keeps our site clean, to avoid irrelevant and incorrect images from staying online.

## This PR

This change both adds delete functionality to the client, but adds a UI component to actually add an image. This deletion occurs both in AWS and the database (as dictated by the backend)

## Screenshots
Screen Shot 2023-11-12 at 2 45 50 PM

## Verification Steps

We manually tested this by:
- Uploading an image on the branch SK.CL.upload-site-image
- Opening AWS > sftt-user-uploads > ... > image folder, ensuring that the image exists
- Clicking the delete button
- Checking that AWS no longer had the image
- Checking that the database site_images folder no longer contains the image

### DIFF
--- a/src/api/protectedApiClient.ts
+++ b/src/api/protectedApiClient.ts
@@ -127,6 +127,9 @@ export interface ProtectedApiClient {
   ) => Promise<void>;
   readonly addSites: (request: AddSitesRequest) => Promise<void>;
   readonly sendEmail: (request: SendEmailRequest) => Promise<void>;
+  readonly deleteImage: (
+      imageId: number,
+  ) => Promise<void>;
   readonly filterSites: (
     params: FilterSitesParams,
   ) => Promise<FilterSitesResponse>;
@@ -197,6 +200,8 @@ export const ParameterizedApiRoutes = {
   UPDATE_SITE: (siteId: number): string => `${baseSiteRoute}${siteId}/update`,
   NAME_SITE_ENTRY: (siteId: number): string =>
     `${baseSiteRoute}${siteId}/name_entry`,
+  DELETE_IMAGE: (imageId: number): string =>
+      `${baseSiteRoute}site_image/${imageId}`,
 };
 
 export const ParameterizedAdminApiRoutes = {
@@ -556,6 +561,12 @@ const sendEmail = (request: SendEmailRequest): Promise<void> => {
   );
 };
 
+const deleteImage = (imageId: number): Promise<void> => {
+  return AppAxiosInstance.delete(
+    ParameterizedApiRoutes.DELETE_IMAGE(imageId),
+  ).then((res) => res.data);
+};
+
 const filterSites = (
   params: FilterSitesParams,
 ): Promise<FilterSitesResponse> => {
@@ -611,6 +622,7 @@ const Client: ProtectedApiClient = Object.freeze({
   nameSiteEntry,
   addSites,
   sendEmail,
+  deleteImage,
   filterSites,
 });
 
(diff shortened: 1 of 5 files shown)


src/api/protectedApiClient.ts
```diff
@@ -556,6 +559,12 @@ const sendEmail = (request: SendEmailRequest): Promise<void> => {
   );
 };
 
+const deleteImage = (imageId: number): Promise<void> => {
+  return AppAxiosInstance.delete(
+    ParameterizedApiRoutes.DELETE_IMAGE(imageId),
+  ).then((res) => res.data);
+};
+
```
huang0h, line 567: ⛏️ Could we add a couple tests for the new route here in `protectedApiClient.test.ts`? These tests are mostly just here as consistency checks to make sure we don't accidentally break a route in the future.

src/components/treePage/siteImageCarousel.tsx
```diff
@@ -87,11 +112,31 @@ export const SiteImageCarousel: React.FC = () => {
...
+                  t('site_image.no_upload_date')}
+              </div>
+              <div>
+                <DeleteSiteImageButton
+                  type="primary"
+                  onClick={() => setShowDeleteForm(!showDeleteForm)}
+                >
+                  Delete
+                </DeleteSiteImageButton>
+              </div>
```
huang0h, line 127: ⛏️ Can we make it so this button only shows up for admins and the user who uploaded the image? The delete site image route is only callable by admins and the image's uploader - we probably shouldn't give other users this option if it won't work for them. 
The image uploader's ID is available from the image, and we can get it the same way we get the image's uploader username or upload date (this field might need to be added to the `SiteEntryImage` type)

src/components/confirmationModal/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+interface ConfirmationModelProps {
+  visible: boolean;
+  onOk: () => void;
+  onCancel: () => void;
+  title: string;
+  confirmationMessage: string;
+  onConfirm: () => void;
+}
+
+const ConfirmationModel: React.FC<ConfirmationModelProps> = ({
```
huang0h, line 25: 🔧 Quick nitpick - did you mean "Modal", not "Model"?

src/components/confirmationModal/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  visible,
+  onOk,
+  onCancel,
+  confirmationMessage,
+  title,
+  onConfirm,
+}) => {
+  return (
```
huang0h, line 33: 💯

huang0h (commented):
Awesome work guys!

PR #319, opened 2024-01-28
Ot.language options

### LINKED ISSUE #234: Add translation drop down to navigation bar
**Block Status**
Blocked by #233

**Why**
- Allows the user to translate the page at the click of a button (easily accessible)

**What to do**
- Create a dropdown with available languages (?) 
- Add it to the left hand side of the sign up button
- Reference: https://react.i18next.com/legacy-v9/step-by-step-guide#d-let-the-user-toggle-the-language

CerberusLatrans:
## Checklist

- [x] 1. Run `yarn run check`
- [x] 2. Run `yarn run test`

## Why

Resolves #234

Provides a dropdown button for users to switch languages (if available)

## This PR

removed FlexBar component from mobileNavBar, replaced with Flex. There is still some weird alignment issue with the button, so that needs to be fixed

## Screenshots
Screenshot 2024-01-28 at 3 56 13 PM

Screenshot 2024-01-28 at 3 55 19 PM

## Verification Steps

Ran the frontend, changed languages successfully

### DIFF
--- a/src/components/navBar/translationDropdown.tsx
+++ b/src/components/navBar/translationDropdown.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import { Dropdown, MenuProps } from 'antd';
 import i18n from 'i18next';
+import { GlobalOutlined } from '@ant-design/icons';
+import styled from 'styled-components';
 
 const items: MenuProps['items'] = [
   {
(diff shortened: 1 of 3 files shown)


huang0h (approved):
lgtm!

src/components/navBar/translationDropdown.tsx
```diff
@@ -0,0 +1,37 @@
...
+const items: MenuProps['items'] = [
+  {
+    key: 'en',
+    label: 'English',
+  },
+  {
+    key: 'es',
+    label: 'Español',
+  },
+];
```
SurabhiKeesara, line 16: Will we be easily able to add other languages to this array? Does this just use the i18 language library/autotranslate?

SurabhiKeesara (approved):
This looks great for English and Spanish, just refer to the question about other languages

src/components/navBar/translationDropdown.tsx
huang0h: Yep! As long we have the translations for other languages, we can just add their key and appropriate label for languages here. We'll just need to make sure that the key we specify here matches what we use in `src/i18n/i18n.ts` where we map languages to translation files.

PR #324, opened 2024-02-06
Emailer: preview HTML content before sending

### LINKED ISSUE #275: Emailer: Preview email content before sending
**Why**
- Admins should be able to see what their email will look like before sending, as currently they only see the raw HTML content of the email

**What to do**

- Add a control to toggle between editing the email content and viewing a preview (consider using the [AntD switch component](https://ant.design/components/switch))
- Using [DOMPurify](https://github.com/cure53/DOMPurify) (you'll have to add it to the repo), display the email HTML content on the page (see [this page on using `dangerouslySetHTML`](https://blog.logrocket.com/using-dangerouslysetinnerhtml-in-a-react-application/) as a guide)

huang0h:
## Checklist

- [x] 1. Run `yarn run check`
- [x] 2. Run `yarn run test`

## Why

Resolves #275 

On the emailer page, add a toggle to preview email HTML before sending

## This PR

- Add DOMpurify as a dependency to sanitize HTML
- Add a switch to toggle between the raw HTML editor and email previewing 

## Screenshots

Raw view:

image

Preview view:

image

## Verification Steps

- Verified the switch toggles between raw and preview views
- Verified HTML content is sanitized in the preview view

### DIFF
--- a/src/components/forms/sendEmailForm/index.tsx
+++ b/src/components/forms/sendEmailForm/index.tsx
@@ -47,7 +47,7 @@ const SendEmailForm: React.FC<SendEmailFormProps> = ({ emails }) => {
   const [bodyContent, setBodyContent] = useState<string>('');
   const [sanitizedBodyContent, setSanitizedBodyContent] = useState<string>('');
 
-  const switchView = (isShowPreview: boolean) => {
+  const togglePreview = (isShowPreview: boolean) => {
     setShowPreview(isShowPreview);
     isShowPreview && setSanitizedBodyContent(DOMPurify.sanitize(bodyContent));
   };
@@ -89,7 +89,7 @@ const SendEmailForm: React.FC<SendEmailFormProps> = ({ emails }) => {
         <Input placeholder={t('subject_placeholder')} />
       </Form.Item>
       <PreviewSwitch
-        onChange={switchView}
+        onChange={togglePreview}
         checkedChildren={t('preview.preview')}
         unCheckedChildren={t('preview.raw')}
       />
--- a/src/components/forms/sendEmailForm/index.tsx
+++ b/src/components/forms/sendEmailForm/index.tsx
@@ -49,7 +49,7 @@ const SendEmailForm: React.FC<SendEmailFormProps> = ({ emails }) => {
 
   const togglePreview = (isShowPreview: boolean) => {
     setShowPreview(isShowPreview);
-    isShowPreview && setSanitizedBodyContent(DOMPurify.sanitize(bodyContent));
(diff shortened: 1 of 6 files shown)


SurabhiKeesara (approved):
This looks good. To follow up on our conversation, I think that this is good for users who want to write custom HTML. However, we should use this in conjunction with something that just inserts some provided text into the body of a pre-selected template.

src/components/forms/sendEmailForm/index.tsx
```diff
@@ -1,53 +1,112 @@
...
+        unCheckedChildren={t('preview.raw')}
+      />
       <Form.Item
         name="emailBody"
-        rules={requiredRule('The email body is required')}
+        rules={requiredRule(t('body_required'))}
+        hidden={showPreview}
       >
-        <Input.TextArea rows={6} placeholder="Email Body" />
+        <Input.TextArea rows={8} placeholder={t('body_placeholder')} />
```
CiciLing, line 101: What's the reason to set the row to 8? if the user has less than 8 lines, would it just show a bunch of white spaces?

src/components/forms/sendEmailForm/index.tsx
```diff
@@ -1,53 +1,112 @@
...
   };
 
   return (
-    <Form name="sendEmail" form={sendEmailForm} onFinish={onFinishSendEmail}>
+    <Form
+      name="sendEmail"
+      form={sendEmailForm}
+      onFinish={onFinishSendEmail}
+      onValuesChange={(changedValues, _) => {
+        if (changedValues.emailBody !== undefined)
```
CiciLing, line 81: if the emailBody is undefined, would we want to prompt the user to do anything? or just ignore until they input something that's valid.

CerberusLatrans (approved):
Good job abstracting the messages using the translation functionality. I Like!

CiciLing (commented):
LGTM! Just have a couple questions on certain implementation.

src/components/forms/sendEmailForm/index.tsx
huang0h: Yep, it's just making the textbox for the the email body a little bigger (see pic). No real important reason behind this, I just thought it'd be nice to have a little more space to view the email HTML (before/after):

image


image

src/components/forms/sendEmailForm/index.tsx
huang0h: Good question! In this case, `emailBody` would be undefined only when the user edits a field that's not the body, since `changedValues` only contains fields that were modified when the user inputs something. This check is just to make sure we don't accidentally clear the body when the user modifies the subject.

PR #327, opened 2024-02-28
Sk.add map to tree page

### LINKED ISSUE #325: Add Map to Tree Info Page
**Why**
- currently on the 'more info' page or looking into the tree site specific tag, we don't know where the tree is based off of that page
- standard users will want to pin point the tree location upon clicking on that specific tree

**What to do**
- consider the edit site page for admins, there is a map there of where the site is
- i.e.
Image
- bring this map into the page for the tree page (pull in location info from the site the tree is on)

SurabhiKeesara:
## Checklist

- [X] 1. Run `yarn run check`
- [X] 2. Run `yarn run test`

## Why

Resolves #325

This change adds a map to the tree info page, allowing users to see where the tree exists relative to other trees and the entire map of Boston

## This PR

DUPLICATED IN https://github.com/Code-4-Community/speak-for-the-trees-frontend/pull/337

Changes include
- Adding a map height parameter to make the height of the map component specifiable depending on where it is called from
- Adding sites and neighborhoods fields to tree props so that the map component can display the entire map of all other sites
- Define a state for the map marker and position
- Add the actual map component to a tree page

## Screenshots

Screenshot 2024-02-28 at 4 23 03 PM
Screenshot 2024-02-28 at 4 22 55 PM

## Verification Steps

To verify that this works
- Select a site from the home page
- Select 'More Info', open in a new tab
- Note that a map exists on the more info page
- Ensure that the map position on the tree page matches the tree position on the home page

### DIFF
--- a/src/components/mapComponents/maps/mapWithPopup/index.tsx
+++ b/src/components/mapComponents/maps/mapWithPopup/index.tsx
@@ -29,8 +29,8 @@ const StyledSearch = styled(Input.Search)`
   }
 `;
 
-const MapDiv = styled.div`
-  height: 100%;
+const MapDiv = styled.div<{ height: string }>`
+  height: ${(props) => props.height || 0};
 `;
 
 interface MapWithPopupProps {
@@ -40,6 +40,7 @@ interface MapWithPopupProps {
   readonly lng: number;
   readonly initMap: (mapData: InitMapData) => ReturnMapData;
   readonly defaultActiveTree: BasicTreeInfo;
+  readonly mapHeight?: string;
 }
 
 let map: google.maps.Map;
@@ -51,6 +52,7 @@ const MapWithPopup: React.FC<PropsWithChildren<MapWithPopupProps>> = ({
   lat,
   lng,
   initMap,
+  mapHeight = '100%',
   defaultActiveTree,
   children,
 }) => {
@@ -222,7 +224,7 @@ const MapWithPopup: React.FC<PropsWithChildren<MapWithPopupProps>> = ({
           onChange={(event) => setSearchInput(event.target.value)}
         />
       </div>
-      <MapDiv id="map" ref={mapRef} />
+      <MapDiv id="map" ref={mapRef} height={mapHeight} />
       <TreePopup treeInfo={activeTreeInfo} popRef={treePopupRef} />
       {children}
     </>
--- a/src/containers/treePage/index.tsx
+++ b/src/containers/treePage/index.tsx
@@ -134,9 +134,6 @@ const TreePage: React.FC<TreeProps> = ({
     nsMode: 'fallback',
   });
 
-  const [editSiteForm] = Form.useForm();
-  const [mapSearchMarker, setMapSearchMarker] = useState<google.maps.Marker>();
-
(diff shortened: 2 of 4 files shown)


src/containers/treePage/index.tsx
```diff
@@ -126,6 +134,9 @@ const TreePage: React.FC<TreeProps> = ({
     nsMode: 'fallback',
   });
 
+  const [editSiteForm] = Form.useForm();
+  const [mapSearchMarker, setMapSearchMarker] = useState<google.maps.Marker>();
+
```
huang0h, line 139: ⛏️ I think we can rid of these variables on the `treePage` - they're for keeping the pin in sync with the lat/long fields of the edit site form (pic) on the edit site page. Since we're not editing the site here, we can keep things simple and not worry about keeping them in sync.

image

src/containers/treePage/index.tsx
```diff
@@ -326,6 +337,20 @@ const TreePage: React.FC<TreeProps> = ({
...
+                              neighborhoods={neighborhoods}
+                              sites={sites}
+                              onMove={(pos: google.maps.LatLng) => {
+                                editSiteForm.setFieldsValue({
+                                  lat: round(pos.lat(), LAT_LNG_PRECISION),
+                                  lng: round(pos.lng(), LAT_LNG_PRECISION),
+                                });
+                              }}
+                              site={siteData.result}
+                              setMarker={setMapSearchMarker}
```
huang0h, line 350: ⛏️ Similar to above - can we replace the values for `onMove` and `setMarker` with an empty lambda, something like `(_) => {}`? (you might need to add the eslint directive `// eslint-disable-next-line @typescript-eslint/no-empty-function` above both lines to pass the linter)

src/components/mapComponents/maps/mapWithPopup/index.tsx
```diff
@@ -29,8 +29,8 @@ const StyledSearch = styled(Input.Search)`
   }
 `;
 
-const MapDiv = styled.div`
-  height: 100%;
+const MapDiv = styled.div<{ height: string }>`
+  height: ${(props) => props.height || 0};
```
huang0h, line 33: 💯

huang0h (changes requested):
Looks great! Just a couple small nitpicks to clean up the code:

huang0h:
> Looks great! Just a couple small nitpicks to clean up the code:

For some reason, when I tried these changes locally the map got stuck in the loading spinner like in the pic below. Did you run into this while testing as well? Not sure if I just messed up my local env and it's an issue on my end.

image

I was able to fix this by adding `dispatch(getMapGeoData());` in the [useEffect hook for this page](https://github.com/Code-4-Community/speak-for-the-trees-frontend/pull/327/files#diff-f04a3895b16734f7838ccc86a9adc050d9f4866b18695bca844714d6e346ee5dR149-R154) (and importing the corresponding function). If nobody else runs into this, then I think you can just leave this out.

huang0h (approved):
lgtm! 👍 not sure what's up with the map display - let's see if it happens in prod after merging

PR #334, opened 2024-03-17
 Emailer: Load HTML templates from S3

### LINKED ISSUE #259: Emailer: Load HTML templates from S3
**Why**
- Admins should be able to load email templates stored in S3 in the emailer
- Admins should be able to save email templates and store them in S3

**What to do**
- Call the [fetch email templates route](https://docs.c4cneu.com/sftt/sftt-api-spec/#get-email-template-names-admin-only) to populate the dropdown with options
- When a user selects a template, call the [load template route](https://docs.c4cneu.com/sftt/sftt-api-spec/#load-email-template-admin-only) to populate the email body with the template contents

CiciLing:
## Checklist

- [ ] 1. Run `yarn run check`
- [ ] 2. Run `yarn run test`

## Why

Resolves #259 

This allows the user to choose the email template and apply it to the email they're sending. The user can make edits to the template. 

## This PR

We passed in a form instance to sendEmailForm/index.tsx from email/index.tsx. 

## Screenshots
Screenshot 2024-03-17 at 1 02 15 PM (1)

## Verification Steps

Choose the template from the select template dropdown and a template in html should appear in the email body.


src/components/forms/sendEmailForm/index.tsx
```diff
@@ -33,9 +34,14 @@ const EmailPreview = styled.div`
...
   readonly emails: string[];
+  readonly sendEmailForm: FormInstance<SendEmailRequest>;
 }
 
-const SendEmailForm: React.FC<SendEmailFormProps> = ({ emails }) => {
+const SendEmailForm: React.FC<SendEmailFormProps> = ({
+  emails,
+  sendEmailForm,
+}) => {
+  const SendEmailForm: React.FC<SendEmailFormProps> = ({ emails }) => {
```
huang0h, line 44: ⛏️ Looks like you have some errors (from merging) - make sure to get rid of this extra declaration of `SendEmailForm`!

huang0h: Also make sure to remove line 50:
```const [sendEmailForm] = Form.useForm();```
since we're passing the form instance as a prop now

File: src/components/forms/sendEmailForm/index.tsx [hunk 2]
```diff
@@ -98,6 +102,11 @@ const SendEmailForm: React.FC<SendEmailFormProps> = ({ emails }) => {
         rules={requiredRule(t('body_required'))}
         hidden={showPreview}
       >
+        <Input.TextArea
+          name="emailContent"
+          rows={6}
+          placeholder={'Email Body'}
+        />
         <Input.TextArea rows={8} placeholder={t('body_placeholder')} />
```
huang0h, line 110: ⛏️ Can we also remove this extra `TextArea`? There's two text areas here:

image

src/containers/email/index.tsx
```diff
@@ -76,13 +79,37 @@ function neighborhoodToId(neighborhood: Neighborhoods): number {
...
+  const [sendEmailForm] = Form.useForm();
+
+  function fetchTemplateNames() {
+    protectedApiClient
+      .getEmailTemplateNames()
+      .then((res) => {
+        setTemplateNames(res.templates);
+      })
+      .catch((err) => {
+        setTemplateNames([defaultTemplate]);
```
huang0h, line 99: ⛏️ Can we add an error message here in the `catch` (same for `fetchTemplateData`? Something like the `message.error()` call in [this function ](https://github.com/Code-4-Community/speak-for-the-trees-frontend/blob/a9c31c362a83a278cdce864b9c92df18ed6ef61f/src/components/forms/sendEmailForm/index.tsx#L71) would do - if something fails on our end, it's better to explicitly tell the user that an error happened than to fail silently and have them guess at why something's not working

huang0h (changes requested):
Overall looks good! Just have a few small comments

CerberusLatrans:
Screenshot 2024-03-31 at 7 04 15 PM
tested error message when loading empty html

huang0h (approved):
lgtm, nice work! A couple extra notes:
- You might need to just re-run the build checks to get them to pass
- When you merge, I suggest choosing the "squash and merge" option (see pic) from the dropdown - if you're not familiar with this, it'll basically combine all of your commits on this branch into a single commit that gets merged into master. This way the commit history doesn't get clogged up with your commits from while you were working

image

PR #355, opened 2024-05-07
Add User Site Reporting to Tree Page

huang0h:
## Checklist

- [x] 1. Run `yarn run check`
- [x] 2. Run `yarn run test`

## Why

Allows users to make report a site for issues: incorrect information displayed, or inappropriate content (tree name/pictures)
See also: https://github.com/Code-4-Community/speak-for-the-trees-backend-v2/pull/230

## This PR

- Adds the `ReportSiteButton` and `ReportSiteForm` components to allow for site reporting
- Stub `siteData` prop on the tree page
  - Map on page needs to load every site marker even though only a small portion is shown, which dramatically slows down the page - temporary fix until we figure out how to make the map more performant

## Screenshots

Button on page:
image

Form in modal:

image

## Verification Steps

- Verified submitting is disabled when no reason is given
- Verified submitting form calls API route to send report email

### DIFF
--- a/src/api/test/protectedApiClient.test.ts
+++ b/src/api/test/protectedApiClient.test.ts
@@ -2287,7 +2287,7 @@ describe('Admin Protected Client Routes', () => {
       };
 
       nock(BASE_URL)
-        .get(ParameterizedApiRoutes.REPORT_SITE(100))
+        .post(ParameterizedApiRoutes.REPORT_SITE(100))
         .reply(200, response);
 
       const result = await ProtectedApiClient.reportSiteForIssues(100, params);
@@ -2304,7 +2304,7 @@ describe('Admin Protected Client Routes', () => {
       };
 
       nock(BASE_URL)
-        .get(ParameterizedApiRoutes.REPORT_SITE(100))
+        .post(ParameterizedApiRoutes.REPORT_SITE(100))
         .reply(400, response);
 
       const result = await ProtectedApiClient.reportSiteForIssues(
--- a/src/api/test/protectedApiClient.test.ts
+++ b/src/api/test/protectedApiClient.test.ts
@@ -2276,4 +2276,43 @@ describe('Admin Protected Client Routes', () => {
       ).catch((err) => err.response.data);
     });
   });
+
+  describe('reportSiteForIssues', () => {
+    it('makes the right request', async () => {
+      const response = '';
+
+      const params = {
+        reason: 'Tree Bad',
+        description: 'just awful',
+      };
+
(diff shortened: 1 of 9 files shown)


src/api/test/protectedApiClient.test.ts
CerberusLatrans, line 1: TESTING?!?!?!

src/components/forms/reportSiteForm/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  const reasonOptions = [
+    {
+      value: 'Inappropriate Content',
+      label: t('report_site.options.inappropriate'),
+    },
+    {
+      value: 'Incorrect Information',
+      label: t('report_site.options.incorrect'),
+    },
```
CerberusLatrans, line 33: At least from the API spec it doesn't seem like reason is categorical but I can see why it makes sense to simplify it to these two cases. In case someone would want to add more reasons in the future maybe this could be a dropdown instead? This is just an idea but the radio is perfectly fine as is

src/components/reportSiteButton/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+const ReportSiteButton: React.FC<ReportSiteButtonProps> = ({ siteId }) => {
+  const { t } = useTranslation(n(site, ['treePage']), {
+    nsMode: 'fallback',
+  });
+
+  const [showReportModal, setShowReportModal] = useState<boolean>(false);
+  const [reportSiteForm] = Form.useForm();
+
+  const reportSite = useCallback(
```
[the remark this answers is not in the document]
huang0h: Honestly, I can't quite remember why. useCallback is technically an optimization but I really don't think it's needed here (whatever performance gains we get are probably trivial at best) - thanks for calling this out. If you're interested, [this section of the useCallback documentation](https://react.dev/reference/react/useCallback#skipping-re-rendering-of-components) goes in-depth into a good case for using it

src/components/forms/reportSiteForm/index.tsx
huang0h: I like the dropdown idea - like you mentioned it'll definitely scale better with more reasons. I kept the API generic so if we do want to add more reasons, we only have to change the frontend, but we can still limit the users' options here so they don't make reports for random things.

CerberusLatrans: coolio!

CerberusLatrans:
Looks good to me! Nice job bravery!

PR #356, opened 2024-05-09
Ot.save html templates

### LINKED ISSUE #335: Emailer: Save HTML templates to S3
**Why**

- Admins should be able to save email templates and store them in S3

**What to do**
- Add a button to save the current email body to S3 by calling the [add email template route](https://docs.c4cneu.com/sftt/sftt-api-spec/#add-email-template-admin-only)
  - We should prompt the user for a name for the template - could show a modal with text input, or open a sub-form similar to the share site feature (in pic)
- Sanitize template contents when we fetch a template and save a template with DOMPurify (example in `togglePreview()` in the `SendEmailForm` component)

 
image

CerberusLatrans:
## Checklist

- [ ] 1. Run `yarn run check`
- [ ] 2. Run `yarn run test`

## Why

Resolves #335 

Allows admins to name and save their email templates

## This PR
Opens a sub modal next to the send email button. Does not allow an empty name.

## Screenshots
Screenshot 2024-05-08 at 9 26 00 PM
Screenshot 2024-05-08 at 9 27 26 PM
Screenshot 2024-05-08 at 9 28 40 PM

## Verification Steps
After saving, made sure was saved in s3 with correct name and html body. The template appears in the drop down menu right after saving.


src/components/saveMenu/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+  return (
+    <SaveMenuContainer>
+      <NameInput
+        defaultValue={templateName}
+        value={templateName}
+        onChange={(e) => setTemplateName(e.target.value)}
+        placeholder={t('name_template')}
+      />
+      <SaveButton onClick={() => onClickSave(templateName, templateBody)}>
+        <SaveTwoTone twoToneColor={LIGHT_GREEN} />
```
huang0h, line 76: IMO I think we can make the icon bigger to make the button a bit more clear - what do you think? You can add a `font-size` style to the icon to change its size, but it might take some work to get it positioned right, so if you do want to make it bigger and it takes more than like 10 minutes then don't worry about it

huang0h (commented):
Overall looks really good, great work 💯 Just a couple quick comments:

src/components/saveMenu/index.tsx
CerberusLatrans: yeah ur right its kinda tiny lmao

huang0h (approved):

PR #363, opened 2024-06-02
Mn.approve reject modal2

maxn990:
## Checklist

- [ ] 1. Run `yarn run check`
- [ ] 2. Run `yarn run test`

#360


src/components/unapprovedImagesTable/index.tsx
```diff
@@ -55,34 +135,116 @@ function responseToTableData(
...
+          )
+        : [],
+    [fetchData, approvedOrRejectedImageIds],
+  );
+
   const tableData = useMemo(
-    () => (fetchData ? fetchData.map(responseToTableData) : []),
-    [fetchData],
+    () => (filteredData ? filteredData.map(responseToTableData) : []),
+    [filteredData],
```
huang0h, line 162: ❓ What's the reason for putting `filteredData` into its own `useMemo` here? Seems like we only use it to compute `tableData` unless I'm missing something

src/components/imageApprovalModal/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+  async function onClickReject() {
+    const toReject: Promise<void>[] = [];
+    if (!data) {
+      return null;
+    }
+    toReject.push(protectedApiClient.rejectImage(data.key, rejectionReason));
+    Promise.all(toReject).then(() => {
+      props.setApprovedOrRejectedImageIds((prevIds) => [...prevIds, data.key]);
+      close();
```
huang0h, line 117: ⛏️ Can we add some feedback with the `message` antd component (specifically, `message.success()` on success, and `message,error()` on failure) once the API call finishes? That way, the user can know for sure the outcome of their action instead of having to assume  it went through fine.

src/components/imageApprovalModal/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+  async function onClickAccept() {
+    const toApprove: Promise<void>[] = [];
+    if (!data) {
+      return null;
+    }
+    toApprove.push(protectedApiClient.approveImage(data.key));
+    Promise.all(toApprove).then(() => {
+      props.setApprovedOrRejectedImageIds((prevIds) => [...prevIds, data.key]);
+      close();
+    });
```
huang0h, line 130: ⛏️ Same as above - can we add some user feedback here with `message` as well?

src/components/imageApprovalModal/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+          </b>
+        </Col>
+        <Col span={12}>
+          <p style={treeSummaryTextStyle}>{lineItem}</p>
+        </Col>
+      </>
+    );
+  }
+
+  const TreeSummaryDisplay = () => {
```
huang0h, line 148: ⛏️ Right now, if someone selects an image on the list view and then switches to the grid view, it's not super clear which ones are selected:

image

Could we add some visuals to show which images are currently selected? Some things I'm thinking are:
- green border around the image
- something like in the figma where a checkmark is overlaid: (although I worry a little it might get read as "approved" instead of "selected")
image

though if you think of something you feel is better, definitely go for it!

huang0h (commented):
💯 awesome work! A couple small comments about the code, and a couple comments about the UX:

src/components/unapprovedImagesTable/index.tsx
```diff
@@ -55,34 +135,116 @@ function responseToTableData(
...
+    );
+  };
+
+  const openApproveImageModal = (data: FilterImageTableData) => {
+    setApproveImageModalOpen(true);
+    setSelectedImage(data);
+  };
+
+  const GridView = () => {
+    const pageSize = 9; // Number of cards per page
```
CerberusLatrans, line 202: do we want the pageSize to stay constant 9 or scale with viewport size?

src/components/unapprovedImagesTable/index.tsx
```diff
@@ -55,34 +135,116 @@ function responseToTableData(
...
   setSelectedImageIds,
+  useGridView,
+  approvedOrRejectedImageIds,
+  setApprovedOrRejectedImageIds,
 }) => {
+  const filteredData = useMemo(
+    () =>
+      fetchData
+        ? fetchData.filter(
+            (data) => !approvedOrRejectedImageIds.includes(data.imageId),
```
CerberusLatrans, line 154: Just to make sure I understand, whenever someone accepts or rejects an image, after closing the approval box modal the images should disappear from grid view without refreshing?

CerberusLatrans: I might just be missing it, but in table view does approving/rejecting also update this `approvedOrRejectedImageIds` list? If not, then the changes made in table view might not be reflected in grid view if not refreshed.

CerberusLatrans:
Great job on this absolutely behemoth of a ticket guys! very impressive. It was a nightmare to read through

